PR #1, opened 2025-04-29
Fill templating files for frontend 

Vebop:
Add docker, dockerfiles, Jenkinsfile, Readmes. All files inspired from LOVE-frontend as we don't have front end based standards in tssw.


Jenkinsfile
```diff
@@ -0,0 +1,154 @@
...
+    disableConcurrentBuilds(
+        abortPrevious: true,
+    )
+    ]
+)
+
+pipeline {
+  agent any
+  environment {
+    registryCredential = "dockerhub-inriachile"
```
Vebop, line 21: Would this be 'anaconda.org/lsstts'?

Jenkinsfile
sebastian-aranda, line 1: @Vebop sorry I'm just looking into this. So, I didn't remember how outdated the Jenkinsfile of the LOVE-frontend was and the specific configurations it has. We shouldn't base on this one, you can see that this Jenkinsfile has steps to build the image, push documentation and even trigger other builds, as I said that is pretty outdated and I need to adjust that in the LOVE-frontend as well! Our main purpose here is to be able to run tests so we can catch errors before merging changes. The part of image builds is covered by the ts_cycle_build repo, which is what I'll try arrangnig ASAP. One step I still don't see how to incorporate is the documentation, but we can think on that later :+1:. **Please adjust the Jenkinsfile to only run node/yarn tests and remove the rest of steps.**

sebastian-aranda (changes requested):
I need to go over the full review, but leaving a comment you can take a look in the meantime :pray: .

docker/Dockerfile
sebastian-aranda, line 1: Similar to the comment about the Jenkinsfile, let's just remove this Dockerfile. I'll arrange it on the ts_cycle_build repo :+1:.

docker/Dockerfile-dev
sebastian-aranda, line 1: Similar to the previous comment, let's just remove this Dockerfile.

docker/Dockerfile-k8s
sebastian-aranda, line 1: Same as previous comment.

docker/Dockerfile-test
sebastian-aranda, line 1: We can keep this Dockerfile for the Jenkinsfile tests step, but we will need to adjust it as discussed on today's logging meeting about using `vite` instead of `yarn`.

docker/docker-compose.yml
sebastian-aranda, line 1: Similar to previous comments, let's just remove this file.

docker/run-tests.sh
sebastian-aranda, line 1: Let's keep this file, but similar to a previous comment we will use `vitest` instead of `yarn test`.

sebastian-aranda (commented):
Review finished @Vebop , there were several things that weren't really needed, sorry about that :pray:. Please take a look to the comments I left and I can approve this :+1:.

Jenkinsfile
Vebop: That makes sense! I applied those changes.
I just looked for any predefined pipelines that would fit us from the jenkins shared library, but there aren't any there for front end projects, right?

sebastian-aranda (approved):
@Vebop I added a commit with a base React + Vite app. This will allow us to have a better wrap of the first PR (and also a good template to write some documentation for javascript projects!) before starting the app development. Sorry to jump in this way, but I found this time it could be faster that I just added the extra details :pray: .

PR #2, opened 2025-05-06
Update version history.

sebastian-aranda:


Vebop (approved):
Thanks for updating this!

towncrier.toml
Vebop, line 1: I think this doesn't get added to the repository, and needs to be added to the .gitignore

sebastian-aranda: In other repos that is being handled by `ts-pre-commit-config`, which creates the `towncrier.toml` (when set to `true` in the `.ts-pre-commit-config.yaml`) and adds it to the `.gitignore`. However, this project is not using `ts-pre-commit-config` because all hooks are mostly related to python packages, that's part of the efforts of extending the documentation and jenkinsshared_library too. We are setting here explicitly the `.pre-commit-config.yaml` with our custom hooks and adding the `towncrier.toml` for that reason :wink:.

Vebop: Ok that sounds good to me!

PR #8, opened 2025-05-15
DM-50905 Update the nightly digest frontend to fetch data using data source adapters

emanehab99:
Update the nightly digest frontend to fetch data using data source adapters; Almanac, ConsDB and Narrative Log.
Currently the following metrics cards on top of the page retrieve data from the data source adapters:
- The exposures metrics card (expected no of exposures is still TBD)
- Telescope Efficiency metrics card
- Time Loss metrics card

The no of tickets in Jira metrics card is TBD

### DIFF
--- a/src/pages/layout.jsx
+++ b/src/pages/layout.jsx
@@ -9,16 +9,19 @@ import ShutterIcon from "../assets/ShutterIcon.svg";
 import EfficiencyIcon from "../assets/EfficiencyIcon.svg";
 import TimeLossIcon from "../assets/TimeLossIcon.svg";
 import JiraIcon from "../assets/JiraIcon.svg";
+import { calculateEfficiency } from "@/utils/fetchUtils";
 
 export default function Layout({ children }) {
   const [exposures, setExposures] = useState(0);
   const [dayObsStart, setDayObsStart] = useState(new Date(2025, 4, 12));
   const [dayObsEnd, setDayObsEnd] = useState(new Date(2025, 4, 13));
   const [instrument, setInstrument] = useState("LSSTCam");
-  const [efficiency, setEfficiency] = useState(0.0);
-  const [timeLoss, setTimeLoss] = useState(0.0);
-  const [weatherLossPercent, setWeatherLossPercent] = useState(0.0);
-  const [faultLossPercent, setFaultLossPercent] = useState(0.0);
+  const [efficiency, setEfficiency] = useState(0);
+  const [efficiencyText, setEfficiencyText] = useState(`${efficiency} %`);
+  const [timeLoss, setTimeLoss] = useState("0 hours");
+  const [timeLossDetails, setTimeLossDetails] = useState(
+    "(- weather; - fault)",
+  );
 
   const handleStartDateChange = (date) => {
     setDayObsStart(date);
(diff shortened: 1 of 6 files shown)


src/utils/fetchUtils.jsx
```diff
@@ -0,0 +1,9 @@
+const calculateEfficiency = (nightHours, sumExpTime, weatherLoss) => {
```
sebastian-aranda, line 1: In general I'm not strict with docstrings for Javascript, but at least I like to add those to utility functions that might be used in several places, and tests as well. Would you consider doing that in this same PR? If you agree, but preffer to do that later, please create a ticket and add a code comment indicating the number of the ticket :pray:.

src/pages/layout.jsx
```diff
@@ -30,26 +75,67 @@ export default function Layout({ children }) {
...
+    if (loss > 0) {
+      let weatherPercent = (weatherLoss / loss) * 100;
+      let faultPercent = (faultLoss / loss) * 100;
+      setTimeLoss(`${loss} hours`);
+      setTimeLossDetails(
+        `(${weatherPercent}% weather; ${faultPercent}% fault)`,
+      );
+    } else {
+      setTimeLoss("0 hours");
+      setTimeLossDetails("(- weather; - fault)");
```
sebastian-aranda, line 125: I have a few suggestion for this code:

1. You could avoid defining state variables that are calculated from other variables and just use a normal const variable. E.g. instead of having a state variable for `efficiency` you can just define a normal var, i.e. `const efficiency = calculateEfficiency(...)` (outside the useEffect scope and inside the `Layout` component). 
2. Efficiency is currently showing all the time 0 because `nightHours`, `sumExpTime` and `weatherLoss` are 0 at the time `calculateEfficiency` is called (in line 110). That's because the parameters are being changed after the fetch callbacks are returned. Here the suggestion is like the opposite of 1. (but complementary): change `nightHours`, `sumExpTime` and `weatherLoss` to be state variables, which then you can use with the calculated variable defined in point 1.

Please apply the same for other variables in the same category (which are only calculated variables to be passed as props): `efficiencyText`, `timeLoss` and `timeLossDetails`.

sebastian-aranda, line 125: Similar to having a `calculateEfficiency` utility method, maybe is worth it too to have one for calculating time loss?

src/pages/layout.jsx
```diff
@@ -30,26 +75,67 @@ export default function Layout({ children }) {
...
+            "Content-Type": "application/json",
+          },
+        },
+      );
+      if (res.ok) {
+        const data = JSON.parse(await res.text());
+        weatherLoss = data.time_lost_to_weather;
+        faultLoss = data.time_lost_to_faults;
+      } else {
+        console.log("error");
```
sebastian-aranda, line 100: I assume you are planning on improving error logging in future PRs?

src/pages/layout.jsx
```diff
@@ -30,26 +75,67 @@ export default function Layout({ children }) {
...
-        setNooftickets(tix.issues.length);
+        const data = JSON.parse(await res.text());
+        nightHours = data.night_hours;
       } else {
         console.log("error");
       }
     }
 
-    fetchJiraTickets();
+    async function fetchNarrativeLog() {
```
sebastian-aranda, line 85: I commented this in a previous PR, but please remember to move this API calls to an utility class/file. Here is an example of a class we have in the LOVE-frontend to manage API calls: https://github.com/lsst-ts/LOVE-frontend/blob/v6.7.2/love/src/Utils.js#L154.

src/pages/layout.jsx
```diff
@@ -9,19 +9,64 @@ import ShutterIcon from "../assets/ShutterIcon.svg";
...
+          headers: {
+            "Content-Type": "application/json",
+          },
+        },
+      );
+      if (res.ok) {
+        const data = JSON.parse(await res.text());
+        setExposures(data.exposures_count);
+        sumExpTime = data.sum_exposure_time;
+        console.log(sumExpTime);
```
sebastian-aranda, line 61: Please remove this `console.log`.

src/pages/layout.jsx
```diff
@@ -9,19 +9,64 @@ import ShutterIcon from "../assets/ShutterIcon.svg";
...
+
+  const handleInstrumentChange = (inst) => {
+    setInstrument(inst);
+  };
 
   useEffect(() => {
-    async function fetchJiraTickets() {
-      let start = dayObsStart.toISOString().split("T")[0];
-      let end = dayObsEnd.toISOString().split("T")[0];
+    console.log("Effect triggered", dayObsStart, dayObsEnd, instrument);
```
sebastian-aranda, line 35: Please remove this `console.log`.

sebastian-aranda (changes requested):
Good job @emanehab99 , this looks good, but there are a couple things I'm asking you to check before merging please :pray: .

sebastian-aranda:
One last thing I almost forgot :bulb:. We have a tool for managing releases versioning: `towncrier`. For this process we require that every developer creates towncrier fragment indicating the type of change(s) the PR that's about to be merge is introducing. There is a README with instructions in the `doc/news` folder. Please @emanehab99 add those too before merging :pray: .

src/pages/layout.jsx
emanehab99: No question!

sebastian-aranda (approved):
Thanks for sticking through the review @emanehab99 !

PR #11, opened 2025-05-23
Change instrument label to say telescope

alserene:


emanehab99 (approved):

Vebop (approved):
I see that this branch/ticket already has a doc/news, so it looks good for this quick change!

PR #15, opened 2025-05-29
DM-51148 Show the frontend as busy while data is being fetched

emanehab99:
Add loading state for each data source and use them to show skeleton in the metrics cards while data is being fetched

### DIFF
--- a/src/pages/layout.jsx
+++ b/src/pages/layout.jsx
@@ -73,6 +73,9 @@ export default function Layout({ children }) {
         setExposureCount(exposuresNo);
         setSumExpTime(exposureTime);
       })
+      .catch(() => {
+        console.error("Error fetching exposures");
+      })
       .finally(() => {
         setExposuresLoading(false);
       });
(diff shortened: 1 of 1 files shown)


src/pages/layout.jsx
```diff
@@ -46,39 +50,52 @@ export default function Layout({ children }) {
...
+      });
+
+    fetchAlmanac(startDayobs, endDayobs)
+      .then((hours) => {
+        setNightHours(hours);
+        setAlmanacLoading(false);
+      })
+      .catch(() => {
+        setAlmanacLoading(false);
+      });
```
sebastian-aranda, line 88: I think, instead of calling `setAlmanacLoading(false);` twice (in the `then` and `catch` calls), you could just have it one time with a `finally` call:
```suggestion
    fetchAlmanac(startDayobs, endDayobs)
      .then((hours) => {
        setNightHours(hours);
      })
      .finally(() => {
        setAlmanacLoading(false);
      });
```
This can be applied to `fetchExposures`, `fetchAlmanac` and `fetchNarrativeLog`.

sebastian-aranda (approved):
Looks good, thank you @emanehab99 ! This is approved, but I left a couple comments for your consideration.

PR #23, opened 2025-06-23
OSW-581: Display dayobs range on sidebar

alserene:
This PR includes:

- The display of the `dayobs` range (or singular `dayobs`) on the sidebar under the parameters.
- A descriptive note by the `Number of Nights` input explaining that the nights are counted backwards from the selected `dayobs`.
- The creation of a new utils file called `calcUtils.jsx`, for storing utility functions that are non-fetch related. Four of those previously in `fetchUtils.jsx` have been moved to this new file.

single_dayobs

dayobs_range


src/utils/calcUtils.jsx
sebastian-aranda, line 1: I have some feelings with the name of this module `calcUtils`. Two things happen to me:
1. it makes me think if we should also move `getDayobsStr`, `getDatetimeFromDayobsStr`, `getDisplayDateRange` to a different file called like `timeUtils`.
2. it makes me wonder what would I do if I need to add a new function which e.g. parses a custom format, like the response payload from a Jira endpoint. Should I include it here? or a new a file like `jiraUtils`?
I was originally suggesting to rename `calcUtils.jsx` to just `utils.jsx` and include everything that is not fetch in there, but I'm not really sure anymore. What do you think?

sebastian-aranda (approved):
Looks good to me :+1:. I left a comment/doubt for your consideration.

src/utils/calcUtils.jsx
alserene: I share in your feelings! I had such a hard time committing to the name of this file as the 5 functions inside are not really related except by them being _not_ fetch utils...

Maybe the approach to thinking about this should be how we want to group our utils. Is it better to group utils by page/component? Or by data source? Or by function/purpose? I think my preference is to group by function but even then I'm not sure what that would look like... `fetchUtils`, `calcUtils`, `displayUtils`, ... I am not sure 🤔

For now, I'm quite happy to rename this one from `calcUtils.jsx` to `utils.jsx`, and maybe once we have collected a few more functions, we will have a better idea of how they might be usefully grouped.

PR #26, opened 2025-07-07
OSW-644  Allow navigating to other pages and pass URL parameters

emanehab99:
Introduce routing to allow navigating to different pages. Also use URL parameters to hold common state/parameters across different pages.
This is infrastructure work before adding the Data Log page


alserene (approved):
Great work!

PR #32, opened 2025-07-23
OSW-702: Fix external links.

alserene:
As per the ticket discussion, the baseUrl has been set to USDF prod as a constant in the utils.
Testing this, it works fine for local development. Hopefully same in production.


alserene:
**NOTE**: I chose to link to the Exposure Detail notebook _without_ matching `dayobs`, `instrument` or other filters because we offer different controls in the Data Log page to the notebook, and I didn't want to cause confusion around this. For example, the notebook only supports one `dayobs`, whereas the app supports many. _However_, it might be worth _at least_ linking to `instrument`==`LATISS`, and perhaps also linking the `dayobs` (and hopefully the user can figure out there is only one day showing there if they wanted more).

- `instrument` linking will be a very small and easy change.
- `dayobs` will just need a little string manipulation to convert `YYYYMMDD` to `YYYY-MM-DD`.

Let me know if you think we should add either/both of these.

Vebop (approved):
Looks like what we discussed to me. I don't think we need to add filters to the exposure detail times square page, because I think it would be confusing if we were to add only some but not all.

sebastian-aranda (approved):

PR #33, opened 2025-07-23
Tickets/OSW-750

alserene:
An update to the info text at the top of the Data Log page which makes it more clear to users what is happening with the filters and other customisations across/within pages.


emanehab99 (approved):
Nice and clear!

PR #44, opened 2025-08-20
OSW-667: Fix NightSummary header layout when the day selector is not shown.

sebastian-aranda:
Hotfix PR to properly set the `NightSummary` header layout when no obsday selector is present.

### DIFF
--- a/doc/news/OSW-667.bugfix.rst
+++ b/doc/news/OSW-667.bugfix.rst
@@ -1,0 +1,1 @@
+Fix NightSummary header layout when the day selector is not shown.


src/components/NightSummary.jsx
```diff
@@ -158,13 +158,19 @@ function NightSummary({ reports = [], nightreportLoading = false }) {
 
   const reportsContainerRef = useRef(null);
 
+  const showObsDaySelector = !nightreportLoading && availableDays.length > 1;
+
   return (
     <Card className="border-none p-0 bg-stone-800 gap-2">
-      <CardHeader className="grid-cols-3 bg-teal-900 p-4 rounded-sm align-center items-center gap-0">
-        <CardTitle className="text-white font-thin">
+      <CardHeader className="grid-cols-4 bg-teal-900 p-4 rounded-sm align-center items-center gap-0">
```
emanehab99, line 165:
```suggestion
      <CardHeader className={`${
            !showObsDaySelector ? "grid-cols-3": "grid-cols-4"} bg-teal-900 p-4 rounded-sm align-center items-center gap-0`}>
        <CardTitle
          className="text-white font-thin col-span-2"
        >
```

emanehab99 (approved):
Nice work, Sebastian!
I would consider making the title shorter though; just "Night Summary". There is an inevitable case when the screen is smaller, forcing the header to expand vertically.

Screenshot from 2025-08-20 13-55-19
what do you think @alserene ?

alserene:
Replying to @emanehab99, sure, I approve changing the title. I did check and "Night Summary" still overflows, but "Night Reports" and just "Reports" doesn't. I like "Night Reports" best. This could work if we do "Report" (singular) for one dayobs, and "Reports" (plural) for multiple dayobs.

Screenshot 2025-08-20 at 2 57 46 PM

Screenshot 2025-08-20 at 2 59 27 PM

Screenshot 2025-08-20 at 3 02 17 PM

sebastian-aranda:
As @emanehab99 said, there is an inevitable case where the title overflows, I don't think there is a solution for all possible sizes. I don't think though the overflow is a problem? I agree with the "standard" screen size it would be good to avoid the overflow, so I'll make the change to "Night Reports" (singular and plural form).

PR #46, opened 2025-08-26
Add In Progress note to Context Feed page.

Vebop:


src/pages/context-feed.jsx
```diff
@@ -1,37 +1,39 @@
 // import { useRouter } from "@tanstack/react-router";
 import React from "react";
 import { useSearch } from "@tanstack/react-router";
-import { TELESCOPES } from "@/components/parameters";
-import { getDatetimeFromDayobsStr } from "@/utils/utils";
 
 function ContextFeed({ loading = false }) {
-  const { startDayobs, endDayobs, telescope } = useSearch({
+  const { startDayobs, endDayobs } = useSearch({
     from: "__root__",
```
alserene, line 7: This one I'd like @emanehab99 or @sebastian-aranda to confirm for me because I'm a little hazy on TanStack Query's routing here. I would have used `from: "/context-feed",` here like I did with the plots and data-log, but maybe root is correct if the other pages are children which access the global params of root?

alserene (approved):
Wonderful! 😎

src/pages/context-feed.jsx
emanehab99: `__root__ `or `/context-feed` would work fine here as long as the context feed uses the common global params only (share the same search schema). If more paramaters are needed later and the search schema is different from the base one (like it's the case for `/data-log`), then it has to be `/context-feed`

PR #53, opened 2025-09-03
Update changelog sidebar link to point to main.

Vebop:


emanehab99:
The changes look good to me. I have a question about the release workflow though. Is a new tag generated for both frontend and backend with every release, even if there might be no changes on either?

sebastian-aranda (approved):
❌Thank you Valerie, looks good :+1:.❌
Edit: I just took a closer look to Eman's comment, the `packageJson.version` is only for the Frontend, sorry I didn't notice the backend was here too... That complicate things a bit as we don't have direct reference on the Frontend about the Backend version. Having said this, and apologizing one more time (without dismissing though all what I explained to you @Vebop :sweat_smile: ), please move things back to `main` for now :pray: . I'll create a ticket to have a version endpoint for the backend API, something like: https://usdf-rsp-dev.slac.stanford.edu/narrativelog/version.

sebastian-aranda (changes requested):
Please roll back to main :pray:.

sebastian-aranda (approved):
Thank you @Vebop !

PR #56, opened 2025-09-16
Tickets/OSW-666

alserene:
PR for the Context Feed.

Please merge the related backend PR first - [#75](https://github.com/lsst-ts/ts_logging_and_reporting/pull/75) :)

Notes:

- If deemed suitable, this should be released as an alpha version for the 17th Sept deployment, in order to collect feedback asap.
- There is a long list of features that could still be implemented. These will be documented in the [OSW-666](https://rubinobs.atlassian.net/browse/OSW-666) ticket. The most important, and at the top of my list, is allowing for colourblind friendly themes.
- Commits will be cleaned up once reviews are completed.


src/components/ContextFeedTable.jsx
```diff
@@ -0,0 +1,544 @@
...
+                            const headerContent = (
+                              <>
+                                {flexRender(
+                                  header.column.columnDef.header,
+                                  header.getContext(),
+                                )}
+                                {/* Active sorting icons */}
+                                {header.column.getIsSorted() === "asc" && " 🔼"}
+                                {header.column.getIsSorted() === "desc" &&
+                                  " 🔽"}
```
Vebop, line 326: it's probably linting, but why isn't the `" v"}` on the same line like the up arrow one ?

src/components/ContextFeedTable.jsx
```diff
@@ -0,0 +1,544 @@
...
+                                  }}
+                                >
+                                  {(() => {
+                                    const currentSort =
+                                      header.column.getIsSorted();
+                                    if (currentSort === false)
+                                      return "Sort by asc.";
+                                    if (currentSort === "asc")
+                                      return "Sort by desc.";
+                                    if (currentSort === "desc") return "Unsort";
```
Vebop, line 384: It looks like there's a double return here, but I think `return "Sort";` is a final one if we don't match any of the other `if`s but, can you fix the spacing if that's the case?

Vebop (approved):
I left a number of comments, but since I'm still new to JS I would prefer someone else's experienced approval, over my noticing of whitespace and small typos.
It feels like a shallow: LGTM

src/components/ContextFeedTable.jsx
alserene: Yep, this is linting. I just tried to move it up to the same line and the linter pushed it back down.

src/pages/ContextFeed.jsx
```diff
@@ -0,0 +1,555 @@
...
+          id: "event_type",
+          value: filterDefaultEventsByTelescope(telescope),
+        },
+      ];
+    });
+  }, [telescope]);
+
+  // Set time ranges from url params.
+  // Must be separate from general url-filter sync to avoid infinite renders.
+  useEffect(() => {
```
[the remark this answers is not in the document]
alserene: Nope. Well, some of it I can pull out (`fullTimeRange` doesn't _actually_ need to be kept in state, so I can move it out, and I will do that, ty), but the rest needs to stay as a `useEffect`. I will try to explain...

We have this code at the top that "subscribes" our component to the URL params:
```
  const { startDayobs, endDayobs, telescope, startTime, endTime } = useSearch({
    from: "/context-feed",
  });
```
Any time a URL param changes, the component gets re-rendered. The key here is what happens on re-render:
- any `const` variables get re-computed.
- `state` variables **DO NOT** get re-computed. They "preserve" their state.
- any `useEffects` dependent on the URL params get re-run.

So, before, I said that `fullTimeRange` doesn't need to be kept in state. That is because it is computed from URL params `startDayobs` and `endDayobs` and it doesn't get modified by anything in this component or by any child component. `selectedTimeRange`, however, does get modified. While it is also _initially_ computed from the URL params, a child component (the Timeline) does modify it. To keep all the components in sync, `selectedTimeRange` is stored in state in the parent component (ContextFeed), and the state, `selectedTimeRange`, and its setter, `setSelectedTimeRange`, are passed to the child components.

And, so, since `selectedTimeRange` needs to be stored in state, it won't be updated when the URL params change. That is why we need a `useEffect`. :)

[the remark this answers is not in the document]
alserene: @nemovc pls correct if I'm wrong 🙏

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+    }
+    return ticks;
+  };
+  const hourlyTicks = generateHourlyTicks(xMinMillis, xMaxMillis, 1);
+
+  //   Dayobs and moon illuminations labels along xAxis
+  //   Show vertical lines at midday dayobs borders
+  //   Show dayobs/illumination labels at midnights (UTC/Chile)
+  const renderDayobsTicks = ({ x, y, payload }) => {
+    const value = payload.value;
```
sebastian-aranda, line 102:
```suggestion
    const { value } = payload;
```

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+
+      return (
+        <>
+          <text
+            x={x}
+            y={y + DIST_BELOW_xAXIS}
+            fontSize={LABEL_TEXT_SIZE}
+            textAnchor="middle"
+            fill="grey"
+            style={{ WebkitUserSelect: "none" }}
```
sebastian-aranda, line 138: Shouldn't `userSelect` be also defined here? I asked GH copilot and this is what he said (of course don't know if that's totally true):

Image

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+
+  const PLOT_HEIGHT = 250;
+  const PLOT_LABEL_HEIGHT = 20; // height of the xAxis label
+  // --------------------------------------------------------
+
+  // Timeline ===============================================
+  return (
+    <ResponsiveContainer
+      title="Time Window Selector"
+      config={{}}
```
sebastian-aranda, line 162: Why is this receiving an empty object? is it really necessary if the component will always have that `{}`?

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+              yAxisId="0"
+              label={{
+                value: `${millisToHHmm(twi)}`,
+                position: "bottom",
+                fill: "white",
+                dy: 5,
+                fontSize: 16,
+                fontWeight: 100,
+                letterSpacing: 0.5,
+                style: { userSelect: "none" },
```
sebastian-aranda, line 229: similar to a previous comment, would we need both `userSelect` and `WebkitUserSelect` here?

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+          .filter(
+            (info) =>
+              info.displayIndex != null &&
+              info.label !== "AUTOLOG" &&
+              info.label !== "Narrative Log",
+          )
+          .map((info) => {
+            const displayIndex = info.displayIndex;
+            const label = info.label;
+            const color = info.color;
```
sebastian-aranda, line 254:
```suggestion
            const { displayIndex, label, color } = info;
```

src/components/ContextFeedTimeline.jsx
```diff
@@ -0,0 +1,319 @@
...
+        ) : null}
+        {/* Selection rectangle shown during active highlighting */}
+        {refAreaLeft && refAreaRight ? (
+          <ReferenceArea
+            x1={refAreaLeft}
+            x2={refAreaRight}
+            fillOpacity={0.3}
+            fill="pink"
+          />
+        ) : null}
```
sebastian-aranda, line 313:
```suggestion
        {refAreaLeft && refAreaRight && (
          <ReferenceArea
            x1={refAreaLeft}
            x2={refAreaRight}
            fillOpacity={0.3}
            fill="pink"
          />
        )}
```
I was going to suggest the same [logical AND conditional rendering approach](https://react.dev/learn/conditional-rendering#logical-and-operator-), but I stepped back because `selectedMinMillis && selectedMaxMillis` seem to be numeric so could lead to the pitfall explained on the docs.

src/components/EditableDateTimeInput.jsx
sebastian-aranda, line 1: I think this one would be benefited of some refactoring in base to: https://react.dev/learn/you-might-not-need-an-effect, specifically by using what's detailed on [Resetting all state when a prop changes](https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes).

src/utils/timeUtils.jsx
```diff
@@ -44,6 +54,34 @@ const getDayobsEndTAI = (dayobsStr) => {
...
+ * Each dayobs ends at midday UTC the next day, so we add a day and return that midday.
+ *
+ * @param {string} dayobsStr - The dayobs date string in "yyyyLLdd" format (e.g., "20250827").
+ * @returns {DateTime} A Luxon DateTime object in UTC at 11:59:00 next day.
+ */
+const getDayobsEndUTC = (dayobsStr) => {
+  const dayobsDate = DateTime.fromFormat(dayobsStr, "yyyyLLdd", {
+    zone: "utc",
+  });
+  return dayobsDate.plus({ days: 1 }).set({ hour: 11, minute: 59, second: 0 });
```
sebastian-aranda, line 82: Just to be strict, maybe set at `11:59:59` instead?

src/utils/timeUtils.jsx
sebastian-aranda, line 1: Would you add unit tests for these new utilities 🙏 ?

src/pages/ContextFeed.jsx
```diff
@@ -0,0 +1,556 @@
...
+    // The dependencies for this useEffect() hook.
+    // If any of these change, this hook will run.
+  }, [startDayobs, endDayobs, telescope]);
+
+  // Filter data based on selected time range
+  // and the event types selected by checkboxes
+  const filteredData = contextFeedData.filter(
+    (entry) =>
+      entry.event_time_dt >= selectedTimeRange[0] &&
+      entry.event_time_dt <= selectedTimeRange[1], // &&
```
sebastian-aranda, line 358:
```suggestion
      entry.event_time_dt <= selectedTimeRange[1],
```

src/components/ContextFeedTimeline.jsx
alserene: `selectedMinMillis`/`selectedMaxMillis` and `refAreaLeft`/`refAreaRight` are both numeric, but I don't see how they can ever be `0`. They're either some large integer like `1753259328477` or they're `null`. Buuuut, just in case, I've added the `> 0` checks like the docs suggest.

PR #63, opened 2025-10-02
OSW-1047: Fix Data Log bug with median psf

alserene:
The fix was a missing `_` in the field name.


emanehab99 (approved):
LGTM, thanks Alice!

Vebop (approved):
good call moving the calculation up out of the return 👍

sebastian-aranda (approved):

PR #64, opened 2025-10-03
OSW-1119: Context Feed: adapt code after rubin_nights changes made

alserene:
This work branched from [OSW-666](https://github.com/lsst-ts/ts_logging_frontend/tree/tickets/OSW-666) (the main Context Feed branch that has an open/approved PR and is awaiting some backend package situation to resolve before being merged).

I'm not sure if this PR should be merged into OSW-666 or into develop, but I went with develop. If it is wrong, just close this one and make a new one, I guess?

:)


Vebop (approved):
Thanks for getting this updated so quickly!

PR #65, opened 2025-10-06
OSW-764: Data Log: add AuxTel support

alserene:
This PR (along with the backend PR) involves a few changes to get AuxTel supported in the Data Log:
- Column definitions have been split into common columns and those specific to each telescope.
- The `RubinTV` links have been adapted for both telescopes.
- Due to the backend bugfix, some returned rows have `dayobs` values of `NaN`, which break the `Plots` page. I've added a little check to fix this.

Other misc changes:
- Default visible columns and column ordering has been moved to `DataLogColumns`.
- Files have been renamed to match standard conventions.

Note: there is a backend PR ([#80](https://github.com/lsst-ts/ts_logging_and_reporting/pull/80)) that needs to be merged before this one.


src/components/dataLogColumns.jsx
```diff
@@ -234,66 +200,253 @@ export const dataLogColumns = [
...
+          exposureName={row.original["exposure_name"]}
+        />
+      ),
+      size: 100,
+      filterType: null,
+      meta: {
+        tooltip: "Link to calibrated exposure in RubinTV. Opens in a new tab.",
+      },
+    }),
+    columnHelper.accessor("exposure_id", {
```
Vebop, line 227: I am concerned that `exposure_id`, `seq_num`, `day_obs`, `air_temp` are not in `commonColumns`. 
Can you explain that, and also why RubinTVLink is displayed in both? This one I could more understand if it has to be calculated differently to link to a different exposure I guess.

src/components/dataLogColumns.jsx
```diff
@@ -234,66 +200,253 @@ export const dataLogColumns = [
...
+    columnHelper.accessor("sky_bg_median", {
+      header: "Sky Brightness",
+      cell: (info) => formatCellValue(info.getValue()),
+      size: 140,
+      filterType: "number-range",
+      meta: {
+        tooltip: "Average sky background (median across all detectors).",
+      },
+    }),
+    columnHelper.accessor("zero_point_median", {
```
Vebop, line 280: Also, maybe we should bring this up with Scot, but I found `zero_point` in consdb's LATISS table, rather than `-median`

Vebop (approved):
Left some comments about columns but otherwise seems good!

src/components/dataLogColumns.jsx
alserene: Regarding the `RubinTVLink`: I think I can change this up a bit to make it a common column. I had them separated because the URLs for AuxTel vs Simonyi are different and, since the `telescope` prop can't be accessed in the `DataLogColumnDefinitions` file, I separated the `RubinTVLink` columns to manually pass the telescope names in. However, I think we can infer the telescope from the first two characters of `exposure_id` (similar to how I'm extracting `day_obs` and `seq_num`).

I will do this and push the update today. Thanks!

src/components/RubinTVLink.jsx
```diff
@@ -1,7 +1,30 @@
...
+  seqNum,
+  exposureName,
+}) {
+  // Handle invalid or missing dayObs/seqNum
+  const invalidDayObs = !dayObs || isNaN(Number(dayObs));
+  const invalidSeqNum = !seqNum || isNaN(Number(seqNum));
+
+  if ((invalidDayObs || invalidSeqNum) && exposureName) {
+    // Schema: length=20 chars
+    if (exposureName.length === 20) {
```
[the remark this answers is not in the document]
alserene: This part is actually a fallback in the should-be-very-unlikely case where `seq_num` and/or `day_obs` are invalid. This really shouldn't actually ever run, but there was another bug in the consdb adapter that was causing these fields to be `null` so I added in this fallback. Now that I've identified and fixed the bug, we shouldn't need this, but as a just-in-case, we might want to leave it in. So, if those fields are invalid _and_ `exposure_name` isn't `20` chars, _then_ the link will be broken.

I assume that `exposure_name` will always be `20` characters and of the form `XX_X_yyyyMMdd_123456`, with the `seq_num` portion being length `6` and padded with `0`'s, but we can ask a consdb person (KT?) to confirm, if you'd like?

PR #114, opened 2026-03-26
OSW-1945: Add new env var ACCESS_TOKEN_INT

Vebop:
Adds the environment variable that we use in this back end pr https://github.com/lsst-ts/ts_logging_and_reporting/pull/116


emanehab99 (approved):
Nice work! Thanks Valerie

PR #117, opened 2026-04-10
OSW-1004: Exposure Breakdown hover highlights data in Observing Conditions

nemovc:
Hovering a bar in the Exposure Breakdown applet now highlights the corresponding observations in the Observing Conditions applet.

Also update the hover opacity for the legend band-based hovering to match, and differentiate between dimmed opacities for ZP & seeing (as they are different colors and sizes, they need different opacities to appear to be 'dimmed' by the same amount according to human perception)

### DIFF
--- a/src/components/ExposureBreakdownApplet.jsx
+++ b/src/components/ExposureBreakdownApplet.jsx
@@ -273,7 +273,7 @@ function ExposureBreakdownApplet({
             </PopoverTrigger>
             <PopoverContent className="bg-black text-white text-sm border-yellow-700 w-[300px]">
               This applet displays a breakdown of exposures taken during the
-              night, grouped by a selected field.
+              specified dayobs range, grouped by a selected field.
               <br />
               <br />
               The chart can be configured to show either the{" "}
@@ -289,7 +289,7 @@ function ExposureBreakdownApplet({
               <ul className="list-disc pl-4 mt-1 space-y-1">
                 <li>
                   Hover over a bar to view total and flagged values, and
-                  highlight the corresponding observations in the{" "}
+                  highlight the corresponding exposures in the{" "}
                   <strong>Observing Conditions</strong> chart.
                 </li>
                 <li>
@@ -298,7 +298,8 @@ function ExposureBreakdownApplet({
                   documentation.
                 </li>
                 <li>
-                  Click a bar to open the Data Log, filtered by that group.
+                  Click a bar to open the <strong>Data Log</strong>, filtered by
+                  that group.
                 </li>
                 <li>Scroll to see additional groups if all are not visible.</li>
               </ul>
--- a/src/components/ExposureBreakdownApplet.jsx
+++ b/src/components/ExposureBreakdownApplet.jsx
@@ -58,6 +58,8 @@ function ExposureBreakdownApplet({
   blockLookup = {},
   exposuresLoading = false,
   flagsLoading = false,
+  onBarHover,
+  onBarLeave,
 }) {
   const [plotBy, setPlotBy] = useState(PlotByValues.NUMBER);
   const [groupBy, setGroupBy] = useState(GroupByValues.SCIENCE_PROGRAM);
@@ -158,8 +160,10 @@ function ExposureBreakdownApplet({
               groupKey,
               unflagged: 0,
               flagged: 0,
+              exposureIds: [],
             };
           }
+          aggregatedMap[groupKey].exposureIds.push(String(row.exposure_id));
 
           if (plotBy === PlotByValues.TIME) {
             if (isFlagged) {
@@ -189,6 +193,7 @@ function ExposureBreakdownApplet({
           groupKey: entry.groupKey,
           unflagged: entry.unflagged,
           flagged: entry.flagged,
+          exposureIds: entry.exposureIds,
           totalValue,
         };
       });
@@ -282,7 +287,11 @@ function ExposureBreakdownApplet({
               <br />
               <strong>Tips:</strong>
               <ul className="list-disc pl-4 mt-1 space-y-1">
-                <li>Hover over a bar to view total and flagged values.</li>
+                <li>
+                  Hover over a bar to view total and flagged values, and
+                  highlight the corresponding observations in the{" "}
+                  <strong>Observing Conditions</strong> chart.
+                </li>
                 <li>
                   In <strong>Science Program</strong> view, hover to see the
                   BLOCK description (if available). Linked labels open the BLOCK
@@ -372,6 +381,7 @@ function ExposureBreakdownApplet({
                               if (tooltipState !== null) {
                                 setTooltipState(null);
                                 setHovered(null);
+                                onBarLeave?.();
                               }
                               return;
                             }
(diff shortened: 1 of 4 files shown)


src/components/ExposureBreakdownApplet.jsx
```diff
@@ -282,7 +287,11 @@ function ExposureBreakdownApplet({
               <br />
               <strong>Tips:</strong>
               <ul className="list-disc pl-4 mt-1 space-y-1">
-                <li>Hover over a bar to view total and flagged values.</li>
+                <li>
+                  Hover over a bar to view total and flagged values, and
+                  highlight the corresponding observations in the{" "}
```
alserene: Not sure if this should be "observations"... 🤔 Other options are "exposures", "data, "data points", "values". Potentially ignore this and keep as is... just noting this down.

alserene: If we make "Observing Conditions" `<strong>`, should we also make "Data Log" `<strong>` (third dot point)?

src/components/ObservingConditionsApplet.jsx
```diff
@@ -460,6 +451,53 @@ function ObservingConditionsApplet({
     return [selectedMinMillis, selectedMaxMillis];
   }, [selectedMinMillis, selectedMaxMillis, xMin, xMax]);
 
+  // opacity for each band based on hovering state.
+  // EBA bar hover takes precedence: dims bands with no matching exposures
+  // within the currently visible x range. Otherwise falls back to legend
+  // hover (hide non-hovered bands).
```
alserene: These should be in sentence case pls  :)

The DM guide doesn't have `.js`/`.jsx` code, but we can extrapolate from Python: https://developer.lsst.io/python/style.html#comments

src/components/ObservingConditionsApplet.jsx
```diff
@@ -52,6 +52,11 @@ import {
 // Constants for gap detection
 const GAP_THRESHOLD = 5 * 60 * 1000;
 
+// Opacity applied to dimmed zero-point bands/dots (EBA hover or legend hover)
+const DIMMED_OPACITY_ZP = 0.02;
+// Opacity applied to dimmed seeing dots (EBA hover or legend hover)
```
alserene: This might just be me, but, because I don't usually include the word "applet" when I think about the Exposure Breakdown, it took me a minute to realise that EBA meant Exposure Breakdown applet.

I know I've used the abbreviations CF and DL, so I guess I would _proabably_ refer to this as EB, but I don't know if I ever have. I might be more likely to use "breakdown" or "barchart".

Anyway, the point of my rambling was to say that I didn't intuitively connect EBA with the Exposure Breakdown applet, so I wonder if you could perhaps change "EBA" to "bar" (or something more explicitly obvious). ty :)

src/components/ObservingConditionsApplet.jsx
alserene: See other comment about "EBA".

alserene: Also, I think this comment needs updating, since we are no longer hiding the non-hovered bands when hovering over the legend.

src/components/ObservingConditionsApplet.jsx
```diff
@@ -460,6 +451,53 @@ function ObservingConditionsApplet({
...
+  // EBA bar hover takes precedence: dims bands with no matching exposures
+  // within the currently visible x range. Otherwise falls back to legend
+  // hover (hide non-hovered bands).
+  const opacity = useMemo(() => {
+    if (hoveredExposureIds !== null) {
+      const visibleData = chartData.filter(
+        (d) => d.obs_start_dt >= xDomain[0] && d.obs_start_dt <= xDomain[1],
+      );
+      return Object.fromEntries(
+        ["u", "g", "r", "i", "z", "y"].map((band) => [
```
alserene: Could we perhaps make the set of bands a CONST since it is used a few times? Probably we use these elsewhere in ND... so maybe best to go in utils, if its not already there...

src/components/ObservingConditionsApplet.jsx
```diff
@@ -460,6 +451,53 @@ function ObservingConditionsApplet({
...
+      );
+    }
+    return {
+      u: hoveredBand && hoveredBand !== "u" ? DIMMED_OPACITY_ZP : 1,
+      r: hoveredBand && hoveredBand !== "r" ? DIMMED_OPACITY_ZP : 1,
+      y: hoveredBand && hoveredBand !== "y" ? DIMMED_OPACITY_ZP : 1,
+      i: hoveredBand && hoveredBand !== "i" ? DIMMED_OPACITY_ZP : 1,
+      z: hoveredBand && hoveredBand !== "z" ? DIMMED_OPACITY_ZP : 1,
+      g: hoveredBand && hoveredBand !== "g" ? DIMMED_OPACITY_ZP : 1,
+    };
```
alserene: Could we make this a little more succinct?

Say we import the set of bands (see comment above):
```
const BANDS = ["u", "g", "r", "i", "z", "y"];
```
Then we could do something like:
```
return Object.fromEntries(
  BANDS.map((band) => [
    band,
    hoveredBand && hoveredBand !== band
      ? DIMMED_OPACITY_ZP
      : 1,
  ])
);
```

src/components/ObservingConditionsApplet.jsx
```diff
@@ -460,6 +451,53 @@ function ObservingConditionsApplet({
...
+  }, [hoveredBand, hoveredExposureIds, chartData, xDomain]);
+
+  // Returns the opacity for a single dot.
+  // When an EBA bar is hovered, matching exposures are full opacity and
+  // non-matching are dimmed. Otherwise falls back to band-level opacity
+  // (driven by legend hover).
```
alserene: Not sure what the "falls back to band-level opacity (driven by legend hover)" means... is this a remnant from when the legend hover completely hid non-hovered values?

src/pages/Digest.jsx
```diff
@@ -76,6 +76,13 @@ export default function Digest() {
 
   const [blockLookup, setBlockLookup] = useState({});
 
+  const [hoveredExposureIds, setHoveredExposureIds] = useState(null);
+  const handleEBABarHover = useCallback(
```
alserene: Refer again to previous comments about me not loving the EBA acronym. Can ignore me.

alserene (approved):
It looks so good! Thanks, Owen 😊 I would suggest, if possible, waiting for @emanehab99 to approve, as she's more familiar with this code.

src/components/ExposureBreakdownApplet.jsx
alserene: This applet actually shows both daytime and nighttime, so can you please change this to

```
This applet displays a breakdown of exposures taken during the
specified dayobs range, grouped by a selected field.
```

See discussion thread: https://rubin-obs.slack.com/archives/C07QJMQKJE6/p1777263752318659

emanehab99 (approved):
Amazing work! Thanks Owen!
@alserene has covered the comment I had about using a `const` for bands. Nothing else from me.

src/components/ObservingConditionsApplet.jsx
nemovc: I'll implement it as a module-level const for now; I think changing other parts of the app to use a unified constant is something for another PR

PR #120, opened 2026-04-22
OSW-1305 Add prominent notes on the ND home page if there are no  consdb entries at all or when there are errors for the nights in question

emanehab99:
A new notification banner system replaced the existing toast-based alert system on all pages with a persistent notification banner component. The change addresses the following problems with the previous approach:

- Error toasts were visually noisy and easy to accidentally dismiss. They also show error details which aren’t relevant to users and don’t belong in the frontend
- There is unnecessarily large number of warning, when nothing actually went wrong.
- Neither distinguished between a data availability issue and a fetch failure.

Banner types are now in use across the application (all pages):

- No data: shown when a fetch succeeds but returns no results for the selected parameters. Dismissible by the user. Resets automatically when a new query is triggered. Conveys that the dashboard is working correctly but there is nothing to show for the selected date range.
- Error: visually distinct from no data, shown when one or multiple data sources fail in the same fetch cycle. Consolidates the failures into a single message than isolated upstream problems. Lists failed sources. Sticky. Hides error details
- System Notice: for application-wide notifications. currently used to show the retention policy banner

Banners are sorted in a fixed order — no data first, then errors — and displayed in a stack at the top of the page content area. They reset on every new fetch, so the banner state always reflects the most recent query.


alserene:
Today we have a Jira authentication error, so we got to see the banner in action! It looks great!

One thing I would like changed, if possible, is to display the error message much sooner. I recorded the same error being displayed in dev (banner) and prod (toast), and the toast appears immediately, making it very clear that the `0` displayed in the Jira card is not necessarily true, whereas the banner does not appear until the Visit Map has loaded, leaving the Jira card looking like it has fetched with no problem (this is especially a problem for the Jira card, because we aren't informing the users within the card that there is a problem. We should probably prioritise that ticket).

PROD (toast):
https://github.com/user-attachments/assets/21a4b1f1-ba3d-44f7-85ee-4bc91d54300c

DEV (banner):
https://github.com/user-attachments/assets/7b7d4e18-cb4b-48b1-bdf7-6cc742a0fea6

I will add these videos also to the ticket, since I think it is easier to view on there.

src/components/NotificationBanner.jsx
```diff
@@ -0,0 +1,81 @@
+import { X } from "lucide-react";
+
+const config = {
+  // systemNotice kept for future PR — do not remove
```
alserene, line 4: Maybe add a TODO (with the expected ticket number) to remove this comment, otherwise it might be missed and remain here.

There is another comment elsewhere, presumably referencing the same ticket, so add a TODO there also.

alserene: A wee reminder that all comments should be in sentence case (capital letter at the start, and full stop at the end unless it is a short phrase).

src/hooks/useNotifications.js
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    notifications,
+    processedNotifications,
+    addNotification,
+    removeNotification,
+    clearNotifications,
+  };
+}
+
+// Placeholder for system notices — to be wired up in a later PR
```
alserene: Add TODO with ticket number here.

src/pages/ContextFeed.jsx
```diff
@@ -201,19 +217,30 @@ function ContextFeed() {
...
+              "The table and the timeline will be empty, but you can still view the almanac data (if available).",
+          });
         }
 
         setRubinNightsData(data);
       })
       .catch((err) => {
         // If the error is not caused by the fetch being aborted
-        // then toast the error message.
+        // then notify the error.
```
alserene, line 221: --> "then notify about the error."

src/pages/ContextFeed.jsx
```diff
@@ -201,19 +217,30 @@ function ContextFeed() {
     fetchContextFeedFromRubinNights(startDayobs, endDayobs, abortController)
       .then(([data]) => {
         if (data.length === 0) {
-          toast.warning("No Context Feed entries found in the date range.");
+          addNotification({
+            type: "noData",
+            source: "context-feed",
```
alserene, line 210: Source is probably `rubin-nights`?

src/pages/ContextFeed.jsx
```diff
@@ -201,19 +217,30 @@ function ContextFeed() {
...
+          err,
+        );
         if (!abortController.signal.aborted) {
-          const msg = err?.message;
-          toast.error("Error fetching Context Feed data from Rubin-nights!", {
-            description: msg,
-            duration: Infinity,
+          addNotification({
+            type: "error",
+            source: "context-feed",
```
alserene, line 229: Source is probably `rubin-nights`?

src/pages/ContextFeed.jsx
```diff
@@ -201,19 +217,30 @@ function ContextFeed() {
...
+            title: "Error fetching Context Feed data",
+            description: "An error occurred while fetching context feed data.",
```
alserene: We should be consistent with capitalisations. "Context Feed" is usually capitalised, but I see from the other descriptions, this is where the source name would go... so maybe this should be "rubin nights data"?

src/pages/ContextFeed.jsx
```diff
@@ -257,25 +284,35 @@ function ContextFeed() {
         // Handle partial errors (one of Zephyr/Jira failing)
         if (blocks.errors) {
           Object.entries(blocks.errors).forEach(([source, message]) => {
-            toast.error(
+            addNotification({
+              type: "error",
+              source: `${getBlockSourceLabel(source)}`,
+              title: "Error fetching BLOCK descriptions",
+              description:
+                "An error occurred while fetching context feed data.",
```
alserene: This isn't context feed data. Perhaps add in `${getBlockSourceLabel(source)}` here for the source.

src/pages/ContextFeed.jsx
```diff
@@ -257,25 +284,35 @@ function ContextFeed() {
...
-          toast.error("Error fetching BLOCK descriptions from Zephyr/Jira", {
-            description: msg,
-            duration: Infinity,
+          console.error(
+            "Error fetching BLOCK descriptions from Zephyr/Jira",
+            err,
+          );
+          addNotification({
+            type: "error",
+            source: "block-lookup",
```
alserene: Should this, perhaps, be "Zephyr/Jira" for the source?

src/pages/DataLog.jsx
```diff
@@ -111,9 +121,14 @@ function DataLog() {
...
     setIllumValues([]);
     setMoonValues([]);
+    clearNotifications();
+    toast.dismiss();
 
     fetchAlmanac(startDayobs, queryEndDayobs, abortController)
       .then((almanac) => {
+        // when does this happen? if the date range is in the future and almanac data is not yet available?
+        // should we show a different message for that case?
+        // should we differentiate between no data vs error fetching data for almanac?
```
alserene: Maybe this could occur if the backend silently fails? We only have one test each for almanac endpoint/service.

src/pages/DataLog.jsx
```diff
@@ -251,25 +279,35 @@ function DataLog() {
         // Handle partial errors (one of Zephyr/Jira failing)
         if (blocks.errors) {
           Object.entries(blocks.errors).forEach(([source, message]) => {
-            toast.error(
+            addNotification({
+              type: "error",
+              source: `${getBlockSourceLabel(source)}`,
+              title: "Error fetching BLOCK descriptions",
+              description:
+                "An error occurred while fetching context feed data.",
```
alserene: Same as above comment. This isn't context feed. Maybe use `${getBlockSourceLabel(source)}`.

src/pages/DataLog.jsx
```diff
@@ -251,25 +279,35 @@ function DataLog() {
...
-          toast.error("Error fetching BLOCK lookups from Zephyr/Jira", {
-            description: msg,
-            duration: Infinity,
+          console.error(
+            "Error fetching BLOCK descriptions from Zephyr/Jira",
+            err,
+          );
+          addNotification({
+            type: "error",
+            source: "block-lookup",
```
alserene: Same again: maybe change to Zephyr/Jira for source? `zephyr-jira`?

src/pages/DataLog.jsx
```diff
@@ -308,9 +346,18 @@ function DataLog() {
     });
   }, [dataLogTableData, selectedTimeRange]);
 
+  const displayedNotifications =
+    tableLoading || almanacLoading ? [] : processedNotifications;
+
   return (
     <>
-      <div className="flex flex-col h-screen w-full p-8 gap-4">
+      <div className="flex flex-col h-screen w-full p-8 gap-6">
```
alserene: The `gap-6` here also increases the space between the timeline and table. Maybe, instead, you could add a `pt-2` to the page header/timeline div below?

src/pages/DataLog.jsx
```diff
@@ -161,20 +180,26 @@ function DataLog() {
...
-          const msg = err?.message || "Unknown error";
-          toast.error("Error fetching data from ConsDB!", {
-            description: msg,
-            duration: Infinity,
+          console.error("Error fetching data log entries:", err);
+          addNotification({
+            type: "error",
+            source: "data-log",
+            title: "Data log data unavailable",
+            description: "Error fetching exposure log or data log",
```
alserene: I don't think exposure log is involved here. Data log is only `cons_db.get_exposures()` and `cons_db.get_transformed_efd_data`.

src/pages/Digest.jsx
```diff
@@ -151,16 +170,19 @@ export default function Digest() {
         setExpectedOnSkyExpCount(expectedSumExposures);
         if (expectedSumExposures === 0) {
           toast.warning(
-            "No expected exposures found for the selected date range.",
+            "Expected number of exposures is zero for the selected date range.",
```
alserene: 👌 Maybe just add "The" to the start of the sentence.

src/pages/Digest.jsx
```diff
@@ -178,10 +200,12 @@ export default function Digest() {
...
-          const msg = err?.message;
-          toast.error("Error fetching almanac!", {
-            description: msg,
-            duration: Infinity,
+          console.error("Error fetching almanac:", err);
+          addNotification({
+            type: "error",
+            source: "almanac",
+            title: "Error fetching almanac",
+            description: "An error occurred while fetching almanac.",
```
alserene: maybe add "data" to the end of the sentence?

src/pages/Digest.jsx
```diff
@@ -195,16 +219,15 @@ export default function Digest() {
...
-          const msg = err?.message;
-          toast.error("Error fetching narrative log!", {
-            description: msg,
-            duration: Infinity,
+          console.error("Error fetching narrative log:", err);
+          addNotification({
+            type: "error",
+            source: "narrative-log",
+            title: "Error fetching narrative log",
+            description: "An error occurred while fetching narrative log.",
```
alserene: Either say "the narrative log" or "narrative log data".

src/pages/Digest.jsx
```diff
@@ -329,24 +358,34 @@ export default function Digest() {
         // Handle partial errors (one of Zephyr/Jira failing)
         if (blocks.errors) {
           Object.entries(blocks.errors).forEach(([source, message]) => {
-            toast.error(
+            addNotification({
+              type: "error",
+              source: `${getBlockSourceLabel(source)}`,
+              title: "Error fetching BLOCK descriptions",
+              description:
+                "An error occurred while fetching context feed data.",
```
alserene: Same again, this not context feed.

src/pages/Digest.jsx
```diff
@@ -329,24 +358,34 @@ export default function Digest() {
...
-          toast.error("Error fetching BLOCK descriptions from Zephyr/Jira", {
-            description: msg,
-            duration: Infinity,
+          console.error(
+            "Error fetching BLOCK descriptions from Zephyr/Jira",
+            err,
+          );
+          addNotification({
+            type: "error",
+            source: "block-lookup",
```
alserene: Same again for source.

src/pages/Layout.jsx
```diff
@@ -80,10 +80,15 @@ export default function Layout({ children }) {
...
           onInstrumentChange={handleInstrumentChange}
         />
-        <main className="flex-1 bg-stone-800 overflow-x-hidden">
+        <main className="flex flex-col flex-1 bg-stone-800 overflow-x-hidden">
           {/* Show/Hide Sidebar toggle */}
           <SidebarToggle />
-          <RetentionBanner />
+          <div className="flex flex-col gap-4 px-8">
+            <RetentionBanner />
+            {/* System notices — deferred, see JIRA-XXX
```
alserene: Update ticket number and add a TODO to remove the comment before merge :)

src/pages/Plots.jsx
```diff
@@ -188,18 +196,24 @@ function Plots() {
...
         }
       })
       .catch((err) => {
         if (!abortController.signal.aborted) {
-          toast.error("Error fetching data log!", {
+          addNotification({
+            type: "error",
+            source: "data-log",
+            title: "Error fetching data log",
             description: err?.message || "Unknown error",
```
alserene: So we want error message shown in the notification banners? Should this instead, perhaps, be diverted to console?

src/pages/Plots.jsx
```diff
@@ -225,9 +243,12 @@ function Plots() {
...
         if (!abortController.signal.aborted) {
-          toast.error("Error fetching almanac!", {
-            description: err?.message,
-            duration: Infinity,
+          addNotification({
+            type: "error",
+            source: "almanac",
+            title: "Error fetching almanac",
+            description:
+              err?.message || "An error occurred while fetching almanac.",
```
alserene: Same here about whether we should show error messages.

src/pages/Plots.jsx
```diff
@@ -300,14 +322,19 @@ function Plots() {
...
         </p>
-        <Toaster expand={true} richColors closeButton />
       </div>
     );
   }
 
   return (
     <>
-      <div className="flex flex-col h-screen w-full p-8 gap-4">
+      <div className="flex flex-col h-screen w-full p-8 gap-6">
```
alserene: Same here as for DL, this increases the gap between page elements. Maybe add `mt-2` to header container?

src/pages/VisitMaps.jsx
```diff
@@ -80,7 +99,13 @@ function VisitMaps() {
 
   return (
     <>
-      <div className="flex flex-col w-full p-8 gap-4">
+      <div className="flex flex-col w-full p-8 gap-6">
```
alserene: Same here about the increased gap.

src/pages/ContextFeed.jsx
```diff
@@ -257,25 +284,35 @@ function ContextFeed() {
         // Handle partial errors (one of Zephyr/Jira failing)
         if (blocks.errors) {
           Object.entries(blocks.errors).forEach(([source, message]) => {
-            toast.error(
+            addNotification({
+              type: "error",
+              source: `${getBlockSourceLabel(source)}`,
```
alserene: `getBlockSourceLabel(source)` returns either "Jira" or "Zephyr". Since none of the other `source` values are capitalised, I think you just want `source` instead of using the mapping. There are 2/3 other uses of this map for the `source` value, so make sure to change them, too.

src/pages/DataLog.jsx
```diff
@@ -111,9 +121,14 @@ function DataLog() {
...
+    toast.dismiss();
 
     fetchAlmanac(startDayobs, queryEndDayobs, abortController)
       .then((almanac) => {
+        // when does this happen? if the date range is in the future and almanac data is not yet available?
+        // should we show a different message for that case?
+        // should we differentiate between no data vs error fetching data for almanac?
         if (almanac === null) {
           toast.warning(
             "No almanac data available. Timeline will be displayed without accompanying almanac information.",
```
alserene: Change toast to a `noData` notification.

If this toast is the only remaining toast on Data Log page, then remove the `toast.dismiss();` after this is gone.

src/pages/Digest.jsx
```diff
@@ -267,16 +291,18 @@ export default function Digest() {
     fetchExposureFlags(startDayobs, queryEndDayobs, instrument, abortController)
       .then((flags) => {
         setFlags(flags);
-        if (flags.length === 0) {
-          toast.warning("No exposures flagged for the selected date range.");
-        }
```
alserene, line 279: If we're keeping _some_ toasts, this would be a good candidate, as long as we flipped it around.

```
if (flags.length !== 0) {
    toast.warning("Some exposures have been flagged!");
}
```

Or "Flagged exposures found!"
Or "Flagged exposures found in the selected date range!"

Something like that.

alserene (approved):
Such great work, Eman! It is such a big improvement to our busy toasts 🫶 Thanks for all the hard work!

I did add a LOT of comments (sorry! 🙈), but they're pretty much all related to the notification text fields and wording. No big changes, just small things. Happy to discuss anything 😊

src/components/NotificationBanner.jsx
```diff
@@ -0,0 +1,81 @@
...
+    title: "text-indigo-200",
+    description: "text-indigo-300",
+    meta: "text-indigo-400",
+    dismiss: "text-indigo-400 hover:text-indigo-200 hover:bg-indigo-900",
+  },
+  noData: {
+    dismissible: true,
+    icon: "–",
+    container:
+      "bg-stone-900 border-l-slate-400 border-t-slate-700 border-r-slate-700 border-b-slate-700",
```
nemovc: Incredibly minor nitpick: the border classes can be rewritten as `border-slate-700 border-l-slate-400` to be a little simpler. Isn't tailwinds fun? 

This also obviously applies to the other notification types

src/pages/ContextFeed.jsx
```diff
@@ -329,7 +368,13 @@ function ContextFeed() {
 
   return (
     <>
-      <div className="flex flex-col w-full h-screen p-8 gap-4">
+      <div className="flex flex-col w-full h-screen px-8 pb-8 gap-4">
```
nemovc: This change removes the padding from the top of the page - which is not consistent with all the other pages.

src/components/NotificationBanner.jsx
```diff
@@ -0,0 +1,81 @@
+import { X } from "lucide-react";
+
+const config = {
+  // systemNotice kept for future PR — do not remove
+  systemNotice: {
+    dismissible: true,
+    icon: "i",
```
nemovc: Is there a reason  we're using single characters as icons here? We use the lucide icon pack in other places in the app (including for the X button in this component)

src/components/NotificationBannerStack.jsx
```diff
@@ -0,0 +1,21 @@
...
+  return (
+    <div className="flex flex-col gap-4">
+      {items.map((banner) => (
+        <NotificationBanner
+          key={banner.id || banner.source}
+          type={banner.type}
+          source={banner.source}
+          title={banner.title}
+          description={banner.description}
+          meta={banner.meta}
```
nemovc: This can be written as `{...banner}`

src/components/RetentionBanner.jsx
nemovc, line 1: Should this component still exist? Obviously, it's dealing with slightly different things than the other notifications, but it's still a 'notification' of sorts and even displays in a similar way. Maybe it should be combined? Maybe not.

src/hooks/useNotifications.js
nemovc: My thoughts are just get rid of this commented code. TODOs in the code, when we have JIRA for actually tracking the work are a bit redundant.

src/pages/ContextFeed.jsx
```diff
@@ -501,7 +546,6 @@ function ContextFeed() {
         />
       </div>
       {/* Error / warning / info message pop-ups */}
```
nemovc: Can get rid of this comment line

src/pages/DataLog.jsx
nemovc: Why do we need to change this at all? But if we do make this change, it should be consistent with the other pages.

src/pages/Layout.jsx
nemovc: (see my previous note regarding commented code)

src/pages/Digest.jsx
```diff
@@ -385,9 +424,30 @@ export default function Digest() {
   const [timeLoss, timeLossDetails] = calculateTimeLoss(weatherLoss, faultLoss);
   const newTicketsCount = jiraTickets.filter((tix) => tix.isNew).length;
 
+  const allLoaded =
```
nemovc, line 404: Another minor nit: I feel it's easier to reason about this sort of thing if it's structured 
`const allLoaded = !(exposuresLoading && expectedExposuresLoading ...);` but that may just be my brain and honestly this just works so it's fine.

src/pages/Plots.jsx
```diff
@@ -287,11 +308,12 @@ function Plots() {
   ]);
 
   const loading = dataLogLoading || almanacLoading;
+  const displayedNotifications = loading ? [] : processedNotifications;
 
   // Temporary display message for AuxTel queries
   if (telescope === "AuxTel") {
     return (
-      <div className="flex flex-col w-full p-8 gap-4">
+      <div className="flex flex-col w-full px-8 pb-8 gap-4">
```
nemovc: Just randomly removes the page padding... for the auxtel fallback error page? Odd

src/utils/notifications.js
```diff
@@ -0,0 +1,49 @@
+import { DateTime } from "luxon";
+
+const makeNotificationId = ({ id, source }) =>
+  id ||
+  `${source ?? "notification"}-${Date.now()}-${Math.random()
+    .toString(36)
```
nemovc: uhhh... why are we generating a random ID *and* a unix timestamp here? Seems overkill

src/utils/notifications.js
```diff
@@ -0,0 +1,49 @@
...
+    type,
+    source,
+    title,
+    description,
+    meta: meta ?? defaultMeta,
+    timestamp: time,
+  };
+};
+
+export const mergeNotifications = (notifications) => {
```
nemovc: Perhaps this function could be renamed `mergeErrorNotifications` or have a comment explaining what and why it does

src/hooks/useNotifications.js
```diff
@@ -0,0 +1,47 @@
+import { useMemo, useState } from "react";
+import { createNotification, mergeNotifications } from "@/utils/notifications";
```
nemovc: Is it worth having a separate utils file for these? It's not like they're large and this file itself isn't large; they're also not used anywhere else and are unlikely to be used anywhere else.

src/utils/notifications.js
```diff
@@ -0,0 +1,49 @@
...
+  const rest = notifications.filter((n) => n.type !== "error");
+
+  if (errorNotifications.length === 0) return notifications;
+
+  const failedSources = errorNotifications.map((n) => n.source);
+  const mergedErrorNotification = createNotification({
+    type: "error",
+    source: "multiple",
+    title: "One or more data sources are unavailable",
+    description: `${errorNotifications.length} source(s) failed to respond. Data may be incomplete.`,
```
nemovc: Why not make the `(s)` conditional on `errorNotifications.length > 1`?

src/components/NotificationBannerStack.jsx
```diff
@@ -0,0 +1,21 @@
...
+export function NotificationBannerStack({ notifications, onDismiss }) {
+  const items = notifications ?? [];
+
+  return (
+    <div className="flex flex-col gap-4">
+      {items.map((banner) => (
+        <NotificationBanner
+          key={banner.id || banner.source}
+          type={banner.type}
+          source={banner.source}
```
nemovc: `source` is passed to the `NotificationBanner` but `NotificationBanner` doesnt' accept `source` as a prop

src/utils/notifications.js
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export const mergeNotifications = (notifications) => {
+  const errorNotifications = notifications.filter((n) => n.type === "error");
+  const rest = notifications.filter((n) => n.type !== "error");
+
+  if (errorNotifications.length === 0) return notifications;
+
+  const failedSources = errorNotifications.map((n) => n.source);
+  const mergedErrorNotification = createNotification({
```
nemovc: This clobbers the `title` and `description` fields of the error messages - but when `addNotification` is used throughout the app to add error messages they include `title` and `description` fields which are silently discarded.

nemovc (commented):
I think @alserene has also mentioned in slack but there isn't clarity about what exactly is a `source` in the error messages, since it can be a page, an endpoint, a component etc. The fact that error messages provide a `title` and `description` which then gets discarded is also very unusual.

There's also a couple of inconsistent padding changes which should be reverted

src/pages/Digest.jsx
nemovc: Why are we keeping _some_ toasts?

src/utils/notifications.js
emanehab99: This is leftover from original behavior; consolidating errors if more than 3. I kept it in case this behavior might change in the future. But maybe not

emanehab99: Sorry I misread the comment. I changed the message altogether

src/hooks/useNotifications.js
nemovc, line 1: Very strange pattern to use a hook-like in a separate file to access a context instead of just using `useContext`

src/pages/Plots.jsx
nemovc: This is not resolved

src/utils/utils.js
```diff
@@ -600,6 +600,66 @@ const getBlockSourceLabel = (source) =>
...
+    source: "multiple",
+    title: "One or more data sources are unavailable.",
+    description: `${failedSources.join(
+      ", ",
+    )} failed to load. Data may be incomplete.`,
+  });
+
+  return [...rest, mergedErrorNotification];
+};
+
```
nemovc: These two functions are not really utilities; they are explicitly only for the NotificationContext and are not used anywhere else (and aren't likely to be used anywhere else) and as such should probably live in that file.

src/contexts/NotificationContext.jsx
nemovc, line 1: Also this file should be .js rather than .jsx as it does not contain any jsx syntax

src/hooks/useNotifications.js
emanehab99: The hook wrapper was suggested by copilot for encapsulation and hiding implementation details, so future refactoring could happen internally without having to touch every consumer of the context.
And I can see this is actually the case. I didn't have to change every page while I'm iterating through the implementation.

src/contexts/NotificationContext.jsx
emanehab99: Doesn't `<NotificationContext.Provider>` count as JSX syntax?

PR #121, opened 2026-04-24
OSW-2071: Show static information while loading in MetricsCard component.

sebastian-aranda:
This PR alters the `MetricCards` skeletons loading behavior so that is only applied to not static data, i.e. always show the labels, but no the data that is being loaded (until it is retrieved).


src/components/MetricsCard.jsx
```diff
@@ -30,26 +30,18 @@ export default function MetricsCard({
...
         </div>
         {/* Render dynamic component and static icons */}
-        {loading ? (
-          <Skeleton className="h-8 w-8 rounded-full bg-teal-700" />
-        ) : typeof icon === "string" ? (
-          <img src={icon} alt={label} />
-        ) : (
-          icon
-        )}
+        {typeof icon === "string" ? <img src={icon} alt={label} /> : icon}
```
alserene:
```suggestion
        {typeof icon === "string" ? (
          <img src={icon} alt={label} />
        ) : loading ? (
          <Skeleton className="h-[40px] w-[40px] rounded-full border-8 border-teal-700 bg-transparent" />
        ) : (
          icon
        )}```

Comment (alserene) [reply src/components/MetricsCard.jsx]: Suggested change:

- IF static icon, display straight away,
- ELSE, until the data loads, show a loading skeleton in the shape of a donut to match the donut chart that will be displayed once the data has fetched.

File: src/components/dialog-metrics-card.jsx [hunk 2]
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import MetricsCard from "./metrics-card";
+import MetricsCard from "./MetricsCard";
```
alserene, line 2: 👌

alserene (approved):
Nice one!

src/components/MetricsCard.jsx
sebastian-aranda: Nice!! I totally missed this, thanks for the great suggestion @alserene 🙌 .

PR #122, opened 2026-04-30
OSW-1234 Update Visit maps help text to align with the recent visit map capabilities from the VisitMapBuilder

emanehab99:
This PR refreshes the Visit Maps page and applet UI to better match the rest of the application and make the map behavior easier to understand.

- Reworked the Visit Maps page layout to use shared page components and a cleaner section structure and match other pages
- Added a collapsible guide that explains the two map views, multi-night behavior, and available controls.
- Moved the legend to a separate section and updated it to a more visual version that covers recent-visit highlighting, and band colors using visit-shaped markers that match the plotted footprints.
- Updated the applet info popover to use shorter, clearer explanations and a visual legend, matching the one in the page.
- Removed the unused `planisphereOnly` request parameter from visit-map fetches to match the backend API.

**P.S.** This is to be merged **after** the [backend PR](https://github.com/lsst-ts/ts_logging_and_reporting/pull/121) is merged

### DIFF
--- a/src/pages/VisitMaps.jsx
+++ b/src/pages/VisitMaps.jsx
@@ -117,7 +117,7 @@ function VisitMaps() {
             onDismiss={removeNotification}
           />
         )}
-       
+
         {/* Page Header, legend and collapsible guide */}
         <div className="flex flex-col gap-2">
           {/* Page title + buttons */}
@@ -141,7 +141,7 @@ function VisitMaps() {
                   onClick={() => setGuideVisible((prev) => !prev)}
                   className="bg-amber-400 text-teal-900 font-sm h-6 rounded-md px-2 shadow-[3px_3px_3px_0px_#0d9488] cursor-pointer hover:bg-amber-300 hover:shadow-[4px_4px_8px_0px_#0d9488] transition-all duration-200"
                 >
-                  {guideVisible ? "Hide Guide" : "Show Guide"}
+                  {guideVisible ? "Hide Tips" : "Show Tips"}
                 </Button>
               </>
             }
@@ -149,7 +149,7 @@ function VisitMaps() {
 
           {/* Map Guide */}
           {guideVisible && (
-            <TipsCard title="Guide">
+            <TipsCard title="Tips">
               <div className="flex flex-col gap-6 text-left">
                 {/* Maps */}
                 <div>
@@ -379,7 +379,6 @@ function VisitMaps() {
           </Card>
         </div>
       </div>
-      <Toaster expand={true} richColors closeButton />
     </>
   );
 }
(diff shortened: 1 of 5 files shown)


src/pages/VisitMaps.jsx
```diff
@@ -81,206 +94,265 @@ function VisitMaps() {
...
                 >
-                  RTN-45
-                </a>
-                :
-              </p>
-              <div className="flex flex-wrap justify-center gap-4 items-center pt-4">
-                <div className="flex items-center gap-1">
-                  <span className="w-3 h-3 bg-band-u"></span>
-                  <span>u band</span>
+                  {guideVisible ? "Hide Guide" : "Show Guide"}
```
nemovc: This is inconsistent with the other pages, which call this section `Tips` rather than `Guide`

src/pages/VisitMaps.jsx
```diff
@@ -81,206 +94,265 @@ function VisitMaps() {
...
+                      </span>{" "}
+                      button animates the sliders automatically through time.
+                    </li>
+                  </ul>
+                </div>
+              </div>
+            </TipsCard>
+          )}
+
+          {/* Legend */}
```
nemovc, line 235: I guess it might be difficult to do, but is there a way to have the legend actually be part of the plot? It takes up a huge amount of space here, and it also feels disconnected from the plots themselves.

src/pages/VisitMaps.jsx
```diff
@@ -81,206 +94,265 @@ function VisitMaps() {
...
+                  <span className={`w-4 h-4 bg-band-z ${VISIT_SHAPE}`} />
+                  <span>z</span>
                 </div>
-                <div className="flex items-center gap-1">
-                  <span className="w-3 h-3 bg-band-y"></span>
-                  <span>y band</span>
+                <div className="flex items-center gap-2">
+                  <span className={`w-4 h-4 bg-band-y ${VISIT_SHAPE}`} />
+                  <span>y</span>
                 </div>
```
nemovc: Could this be replaced with a `BANDS.map(b=>...` using the bands constant we've defined elsewhere?

nemovc (approved):
It's definitely an improvement. The legend does take up a huge amount of dead space, I'm not sure if there's a better way to lay it out though

src/pages/VisitMaps.jsx
alserene: Yeah, that was me. I had reasons. They don't seem very salient now, though. Happy for this to be `Tips` for consistency 👍

src/pages/VisitMaps.jsx
alserene: Not easily, I don't think? We'd potentially have to submit a PR to `schedview`.

emanehab99: Or see if there is a way to add more UI elements to the builder

src/pages/VisitMaps.jsx
emanehab99: This didn't work. Tailwind does not evaluate JavaScript variables when it scans for classes. I defined a const for tailwind band classes in PLOT_DEFINTIONS instead

PR #123, opened 2026-05-08
OSW-2119: Hide tips by default

alserene:
The original ticket was for just the Context Feed, but we should have consistency across our pages.
All pages now have tips hidden by default.


emanehab99 (approved):
Looking good. Thanks Alice!

PR #125, opened 2026-05-12
OSW-2240: Fix hover state in exposure breakdown applet

nemovc:
Very strange issue; recharts blocks mouse events from propagating correctly so we have to use a workaround


alserene (approved):
Many thanks for fixes the janks 😎

PR #126, opened 2026-05-14
OSW-2286: Explore echarts migration

nemovc:
Adds a replacement timeline component built using echarts instead of recharts.

Functionally very similar to the old component; with only superficial changes (except for the selection tool):

Event icons are slightly changed to look cleaner, which also provides us a significant performance boost
The selection area is always grey, rather than pink while selecting then grey when inactive.
When the full time range is selected, the selection area disappears, rather than overlaying at full-width

Otherwise, the icons, text, coloring, sizing, padding etc have all been chosen to replicate the old component as closely as possible.

The selection tool is improved and now uses the native echarts `brush` instead of a direct DOM manipulation workaround. This provides the following new features:
- Handles on the selected time range allowing the user to expand/reduce the selection from either end
- The selection itself is draggable, allowing the user to shift the selected area while keeping the size of the selection area the same
- we lose the shift-extension (it's now covered by the selection brush just implicitly having resize handles)


src/pages/ContextFeed.jsx
```diff
@@ -408,11 +413,10 @@ function ContextFeed() {
                     <span className="font-bold">Drag</span> to select a time
                     range (table updates automatically).
                   </li>
+                  <li>Drag the edges of the selection to resize.</li>
                   <li>
-                    <span className="font-bold">Shift + Drag</span> to extend
-                    selection.
+                    The selected time range can also be dragged to reposition.
                   </li>
```
alserene:
```suggestion
                  <li>
                    <span className="font-bold">Drag</span> the selection to reposition.
                  </li>
                  <li>
                    <span className="font-bold">Drag</span> the edges of the selection to resize.
                  </li>
```

src/pages/DataLog.jsx
```diff
@@ -385,11 +385,10 @@ function DataLog() {
                     <span className="font-bold">Drag</span> to select a time
                     range (table updates automatically).
                   </li>
+                  <li>Drag the edges of the selection to resize.</li>
                   <li>
-                    <span className="font-bold">Shift + Drag</span> to extend
-                    selection.
+                    The selected time range can also be dragged to reposition.
                   </li>
```
alserene:
```suggestion
                  <li>
                    <span className="font-bold">Drag</span> the selection to reposition.
                  </li>
                  <li>
                    <span className="font-bold">Drag</span> the edges of the selection to resize.
                  </li>
```

src/pages/Plots.jsx
```diff
@@ -360,9 +360,9 @@ function Plots() {
                   <span className="font-bold">Drag</span> to select a time range
                   (plots update automatically).
                 </li>
+                <li>Drag the edges of the selection to resize.</li>
                 <li>
-                  <span className="font-bold">Shift + Drag</span> to extend
-                  selection.
+                  The selected time range can also be dragged to reposition.
                 </li>
```
alserene:
```suggestion
                  <li>
                    <span className="font-bold">Drag</span> the selection to reposition.
                  </li>
                  <li>
                    <span className="font-bold">Drag</span> the edges of the selection to resize.
                  </li>
```

alserene (approved):
Nice work, Owen!

Two requests, please:

1. Please update the tips as I've suggested (or along those lines, but note the bolding of action words).
2. Please make the timeline width responsive so that it does not overflow off the page on smaller screens (see screenshots).

DEV:
Image
PROD:
Image

PR #127, opened 2026-05-14
OSW-2155: Data-log tests

nemovc:
Add end to end tests for the data-log page. 

Extract some shared tests such as checking that tips are displayed and toggleable.

Also includes some new tests for the routing/URL validation.


src/routes.js
```diff
@@ -180,11 +184,20 @@ function stripUnknownParams(allowedKeys) {
...
 
+// Wrap schemas in plain functions to avoid TanStack Router's Standard Schema
+// (~standard) interface. Zod v3.24+ implements ~standard, but when _parseSync()
+// ever throws an "encountered" error it permanently marks the schema instance as
+// async, causing all subsequent validations to return Promises and TanStack Router
+// to throw "Async validation not supported". Using a plain function forces the
+// function path, which calls schema.parse() directly. ZodError.message is already
+// a JSON string of issues, so SearchParamErrorComponent parses it correctly.
+const parseWith = (schema) => (search) => schema.parse(search);
+
```
alserene, line 195: Thanks for the detailed comment ⭐️

tests/e2e/mocked/shared/almanac.spec.js
```diff
@@ -0,0 +1,88 @@
...
+      moon_rise_time: "2026-01-01 22:00:00",
+      moon_set_time: "2026-01-02 02:00:00",
+      moon_illumination: "50%",
+    },
+  ],
+};
+
+// Color constants (from PLOT_DEFINITIONS.js)
+const TWILIGHT_STROKE = "#0ea5e9";
+const MOON_FILL = "#EAB308";
```
alserene, line 30: Should these not be imported? ...or does this take us down a slippery slope of which UI test elements should be hard-coded vs not?

tests/e2e/mocked/datalog/table.spec.js
alserene, line 1: Can we also test that:

1. `RubinTV` column correctly displays a link.
2. A `Science Program` with matching `BLOCK` data (link/description) correctly shows a link under the `Science Program` column and the description under the `BLOCK Description` column.

alserene (approved):
Nice work, Owen! Thanks for the detailed testing and easy-to-follow organisation of tests 😊

1. One request would be to test that a multi-night Data Log appears as is expected. There are also combinations of data/no-data dayobs that might be good to test. (I think Eman also asked about these for the Plots tests.)
2. You'll have to swap out the warning toast tests to cover their replacement no data/error banners.
3. (I guess also with the eCharts timeline, the timeline tests will need updating then, too.)

tests/e2e/mocked/datalog/table.spec.js
alserene: @nemovc, poke.

alserene:
@nemovc, it seems I don't need to approve this again, so you can merge once you've added a news fragment. There are also a couple of questions/requests I added in my review for you to address pls and thankye. :)

alserene:
Just a note here that the cell expansion bugfix PR ([OSW-2754](https://github.com/lsst-ts/ts_logging_frontend/pull/141)) has some changes that conflict with this PR.

[OSW-2754]: https://rubinobs.atlassian.net/browse/OSW-2754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

tests/e2e/mocked/shared/almanac.spec.js
sebastian-aranda: +1 to Alice's comment. Are we able to import the colors from constants defined inside the package?

tests/e2e/helpers/mock-generators.js
```diff
@@ -48,8 +49,72 @@ export function generateDataLogMock(
...
+ * @param {object} [options] - Same options as generateDataLogMock
+ * @returns {{ data_log: object[] }}
+ */
+export function generateDataLogMockMultiBand(count, options = {}) {
+  const bands = ["y", "g", "r", "i", "z"];
+  return generateDataLogMock(count, {
+    ...options,
+    postProcess: (r, i) => ({
+      ...r,
+      band: bands[i % bands.length],
```
sebastian-aranda, line 73: Nit pick 💡:

```suggestion
      band: bands[(i - 1) % bands.length],
```
As indices are supposed to be 1-based.
I accepting the suggestion, please apply here and elsewhere.

sebastian-aranda (approved):
Amazing work @nemovc 👏 ! this looks great, I just left a couple minor comments for your consideration.

tests/e2e/helpers/mock-generators.js
nemovc: Not sure I understand your intent here - that just changes it from the sequence being `y, g, r, i, z, y, g, ...` to being `g, r, i, z, y, g, r, ...` which doesn't actually change anything

sebastian-aranda: Weird I'm pretty sure I saw a `datalog-multi-band.json` where the first element had `band = g`. In any case, I can explain the situation with the `generateDataLogMockMultiProgram`: when doing the review, after looking at this file I went to check the fixtures, when looking at them my first though was "oh weird, I saw these programs, but I think they started with FM (instead of BF in the first element), **something might be wrong**", then I went back to the file to notice the comment which said the postProcess `i` argument was 1-based. That's why this is a nit pick, just a slight improvement in readability.

PR #130, opened 2026-05-26
OSW-1321: Context Feed + Data Log: make rows in table focusable / linkable

nemovc:
Adds the ability to click to select a single row, and also syncs the selected row state with the URL, allowing it to be shared and for links to specific rows to be generated. If a URL contains a selected row, the table will scroll to display that row on load.

There is ongoing discussion regarding whether and how to support multi-line selection; this PR is mostly to get feedback on the code as it is now and to get it into dev so users can begin using it and providing feedback.


src/components/ui/table.jsx
```diff
@@ -55,7 +55,7 @@ function TableRow({ className, ...props }) {
     <tr
       data-slot="table-row"
       className={cn(
-        "hover:bg-muted/20 data-[state=selected]:bg-muted border-b transition-colors",
+        "hover:bg-muted/10 data-[state=selected]:bg-muted border-b transition-colors",
```
nemovc, line 58: This decreases the brightness of the hover so that we don't need to mess around with changing the color of links to ensure visiblity/accessibility

src/components/ContextFeedColumns.jsx
```diff
@@ -42,14 +42,12 @@ function renderNameCell(info) {
...
-      // Wrap in a dark background for visibility
-      // when row is highlighted.
-      <div className="bg-stone-800 p-1 rounded">
+      <div className="p-1 rounded">
         <a
           href={block.url}
           target="_blank"
           rel="noopener noreferrer"
-          className="text-sky-500 underline hover:text-sky-400"
+          className="text-sky-500 underline hover:text-sky-300"
```
nemovc: Same color as other links

src/components/ContextFeedColumns.jsx
```diff
@@ -138,9 +134,12 @@ function renderDescriptionCell(info) {
...
     <div
       className="relative"
-      onClick={() => toggleExpandedRows(rowId, "description")}
+      onClick={(e) => {
+        e.stopPropagation();
+        toggleExpandedRows(rowId, "description");
+      }}
     >
-      <pre className="whitespace-pre-wrap text-xs bg-stone-900 p-2 rounded cursor-pointer">
+      <pre className="whitespace-pre-wrap text-xs text-white bg-stone-900 p-2 rounded cursor-pointer border border-transparent [tr[data-selected=true]_&]:border-stone-800">
```
nemovc: Add a slight border when selected to increase visibility

src/pages/ContextFeed.jsx
```diff
@@ -85,27 +84,26 @@ function ContextFeed() {
...
+  // Selection state synced with URL
+  const { selectedValue, setSelectedValue } = useSelectionSync({
+    routePath: "/context-feed",
+    selectedKey: "time",
+  });
+
   // Data and loading flags
   const [rubinNightsData, setRubinNightsData] = useState([]);
-  const [rubinNightsDataLoading, setRubinNightsDataLoading] = useState(false);
+  const [rubinNightsDataLoading, setRubinNightsDataLoading] = useState(true);
```
nemovc, line 96: While the initial state is `false`, it immediately changes to `true` and this causes multiple renders; a side effect of which is that the onload scroll doesn't fire correctly.

alserene: Is this one of the causes of multiple renders in other pages?

src/pages/ContextFeed.jsx
```diff
@@ -85,15 +85,21 @@ function ContextFeed() {
   const { selectedTimeRange, setSelectedTimeRange, fullTimeRange } =
     useTimeRangeFromURL("/context-feed");
 
+  // Selection state synced with URL
+  const { selectedValue, setSelectedValue } = useSelectionSync({
+    routePath: "/context-feed",
+    selectedKey: "time",
```
alserene: Can we change the URL param from UTC to unix ms in line with `startTime` and `endTime`?

Then, instead of`selected_time=2026-05-22T08%3A27%3A36.491990%2B00%3A00`
we would have `selected_time=1779438456491`.

alserene:
If a row is selected, and then some section of text in the row is selected (for purposes of copy-paste actions), the row unselects. (Text selection inside an expandable does not unselect row.)

src/components/ContextFeedColumns.jsx
```diff
@@ -42,14 +42,12 @@ function renderNameCell(info) {
   // If available, display BLOCK names as Zephyr/Jira links
   if (block) {
     return (
-      // Wrap in a dark background for visibility
-      // when row is highlighted.
-      <div className="bg-stone-800 p-1 rounded">
+      <div className="p-1 rounded">
```
alserene: Please remove `className="p-1 rounded"` as this is creating nonsense padding now the stone background has been removed (you might be able to remove the `<div>` entirely).

alserene:
Something either to be done in this PR or we can make a ticket for it (and add it as a child of the [accessibility ticket](https://rubinobs.atlassian.net/browse/OSW-2710)) is that the rows are not keyboard focusable/selectable with tab and up/down arrow keys. Currently, the only elements of the table to navigate through are the links. TanStack does not natively support this, so we'd need to configure ourselves.

See [this TanStack GH discussion](https://github.com/TanStack/table/discussions/2752) and [this rival table library's list](https://www.simple-table.com/blog/free-react-tables-accessibility-keyboard-navigation) of what we need to add to TanStack to make it accessible.

src/pages/ContextFeed.jsx
alserene: @nemovc?

alserene (approved):
Yes, I happy. Thank you. Two months later, we got there.

I approve the merging of this PR, but note that I asked a question somewhere about the loading state's initial value (regarding other pages), and I noted something about accessibility, which we'll need to address _at some point..._

src/pages/ContextFeed.jsx
sebastian-aranda: also please match camel case 🙏 : `selected_time` -> `selectedTime`

sebastian-aranda: and `selected_exposure_id` -> `selectedExposureId`

sebastian-aranda: I just saw the snake case format is required in the logic of the `useSelectionSync`... Any possibility of changing the logic or we live with different URL argument formats: camelCase vs snake_case.

doc/news/OSW-1321.feature.rst
```diff
@@ -0,0 +1 @@
+Adds the ability to click to select a single row, and also syncs the selected row state with the URL, allowing it to be shared and for links to specific rows to be generated.
```
sebastian-aranda: _In general_ we use imperative tense (`Adds` -> `Add`) in our Nightly Digest change logs, we have never defined a standard, **we should in our next meeting and document it**, probably in `doc/news/README.rst` 👌 .

src/components/DataTable/DataTable.jsx
```diff
@@ -151,6 +169,9 @@ const DataTable = forwardRef(function DataTable(
               table={table}
               columns={columns}
               isLoading={isLoading}
+              selected={selected}
+              onSelectionChange={onSelectionChange}
+              selectedKey={findSelectedKey(columns)}
```
sebastian-aranda: Please don't set a prop inferred from another one: `selectedKey = fx(columns)`. Instead get it at render time inside `DataTableBody`.

src/components/DataTable/DataTableBody.jsx
```diff
@@ -13,8 +15,37 @@ import { SKELETON_ROW_COUNT } from "./constants";
...
+      !selectedRowRef.current
+    )
+      return;
+    hasScrolled.current = true;
+    selectedRowRef.current.scrollIntoView({
+      behavior: "smooth",
+      block: "center",
+      container: "nearest",
+    });
+  }, [isLoading, selected]);
```
sebastian-aranda: I think there is a chance of race conditions triggering here due to the useEffect call on `isLoading` or `selected` changes, but `selectedRowRef.current` not asigned yet, hence having no scroll effect. I suggest the following change, converting TableRow ref assignment to a callback allows to trigger effects such as scrolling. This way we ensure scrolling is only called when the DOM ref is assigned.
```suggestion
  const scrollToSelectedRow = (node) => {
    if (node && !hasScrolled.current) {
      node.scrollIntoView({
        behavior: "smooth",
        block: "center",
        container: "nearest",
      });
      hasScrolled.current = true;
    }
  }

  const setSelectedRowRef = (node) => {
    if (node) {
      selectedRowRef.current = node;
      scrollToSelectedRow(node);
    }
  };

  useEffect(() => {
    if (isLoading ||selected === null) return;
    scrollToSelectedRow(selectedRowRef.current);
  }, [isLoading, selected]);
```

and then:
```
          <TableRow
            key={row.id}
            ref={isSelected ? setSelectedRowRef : null}
            ...
```

src/components/DataTable/DataTableBody.jsx
sebastian-aranda: Adding this at the root of this file:
```
const MemoTableRow = memo(({ ref, row, table, columns, ...props }) => {
  const isGroupedRow = row.getIsGrouped();
  return (
    <TableRow ref={ref} {...props}>
      {isGroupedRow ? (
        <GroupedRowCell row={row} table={table} columns={columns} />
      ) : (
        <NormalRowCells row={row} />
      )}
    </TableRow>
  );
}, (prevProps, nextProps) => {
  return (
    prevProps.isSelected === nextProps.isSelected &&
    prevProps.row === nextProps.row &&
    prevProps.table === nextProps.table &&
    prevProps.columns === nextProps.columns
  );
});
```
And changing this:

```suggestion
         <MemoTableRow
            key={row.id}
            ref={isSelected ? selectedRowRef : null}
            isSelected={isSelected}
            row={row}
            table={table}
            columns={columns}
            onClick={handleClick}
            data-selected={isSelected ? "true" : undefined}
            className={
              isSelected
                ? "bg-black/40 shadow-[inset_4px_0_0_0_white] border-t-2 border-b-2 border-white"
                : ""
            }
          />
```
Would improve rendering avoiding to re-render every `TableRow` (which props in turn are dependent on `isSelected`) on row selection. Additionally to `isSelected`, in the memo props comparison we check for `row`, `table` and `columns` to make sure to render `TableRow` children content updates.

Every TableRow gets rendered on updating the selected row:
https://github.com/user-attachments/assets/a134c8a5-ce4a-4320-b5ea-e834df266cf3

Only the respective row(s) get rendered:
https://github.com/user-attachments/assets/78612d46-520b-468c-adc9-2d9e7d98b6b8

sebastian-aranda (changes requested):
Nice work @nemovc ! I would approve this PR but I would like you to consider a performance improvement to the DataTableBody component before merging. Please check my comments and request my re-review in any case 👍 .

nemovc:
> Something either to be done in this PR or we can make a ticket for it (and add it as a child of the [accessibility ticket](https://rubinobs.atlassian.net/browse/OSW-2710)) is that the rows are not keyboard focusable/selectable with tab and up/down arrow keys. Currently, the only elements of the table to navigate through are the links. TanStack does not natively support this, so we'd need to configure ourselves.
> 
> See [this TanStack GH discussion](https://github.com/TanStack/table/discussions/2752) and [this rival table library's list](https://www.simple-table.com/blog/free-react-tables-accessibility-keyboard-navigation) of what we need to add to TanStack to make it accessible.

created [ticket ](https://rubinobs.atlassian.net/browse/OSW-2746)

src/components/DataTable/DataTableBody.jsx
nemovc: So the reason this doesn't happen - useEffect runs the tick _after_ the render settles. So, since we're always (well, so long as something is selected in the URL on load) setting `selectedRowRef` to a table row at the same time loading completes (`isLoading` set to `false`) then `selectedRowRef` will always resolve to a table row.

However, I think your solution is cleaner and easier to understand - "when this row is added to the dom, scroll to it" rather than "when this ref is populated, scroll to the row it points to"

src/components/DataTable/DataTableBody.jsx
nemovc: Wow. I'm a donut. This is a huge performance gain, well spotted :)

src/pages/ContextFeed.jsx
nemovc: Nope. That was caused by the other `useTimerangeFromURL` hook double handling state, and was resolved [here](https://github.com/lsst-ts/ts_logging_frontend/pull/132). These double renders also occured on page load (before data load) and were therefor very quick since there wasn't a huge chunk of data to rerender, just the skeletons; however since the old (now superseeded) scroll useEffect would trigger when `isLoading` changes from initially `false` to `true` when the data load starts (as part of the first render), it would trigger on page load, go "oh, can't find a ref to scroll to, but I tried anyway so I'll set 'hasScrolled' to true and not scroll again" and then a second later `isLoading` is changed to false because the data has loaded, but since `hasScrolled` is already true by that point it never scrolled

This is now superseeded by the to the useEffect proposed by @sebastian-aranda but leaving it in since it's cleaner

sebastian-aranda (approved):
Great job @nemovc , thanks for quickly addressing the comments 🙌 !

PR #132, opened 2026-05-30
OSW 2285: Context Feed/Data-log cascading rerenders

nemovc:
The URL sync hooks had an issue where they would cause multiple cascading renders (one would update state and navigate, which caused the other one to rerun even though its state hadn't actually changed, which would also cause the first to rerun and then settle). This was because the `setState` and `navigate` calls were not batching as expected. By removing the `useState` and just deriving state directly from the URL, as well as being more selective about which bits of URL state are watched we bypass this and prevent multiple renders occurring, leading to a dramatic increase in performance.

image

### DIFF
--- a/doc/news/OSW-2285.perf.rst
+++ b/doc/news/OSW-2285.perf.rst
@@ -1,0 +1,1 @@
+Reduced unnecessary re-renders on the Data Log and Context Feed pages.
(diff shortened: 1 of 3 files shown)


nemovc:
Also, while verifying that the changes actually led to an increase in performance and decrease in duplicate renders, I created a set of performance-measuring tests (see screenshot above). It may be worth actually adding some of these to the end to end tests and configuring the CI to give us a warning of some sort if there's a significant degradation against some baseline

alserene (approved):
Wonderful, wonderful. Many thanks 🙏

It is a good idea to add performance tests to our Playwright suite, but if there is still some question of testing/discussing the report generation capabilities of Jenkins with @sebastian-aranda, it might be best to postpone this for a couple of weeks. Maybe create a ticket and add your thoughts there to begin your discussion.

:)

PR #133, opened 2026-06-04
OSW-1475 Replace the interactive visit map applet with a static one 

emanehab99:
This PR replaces the interactive visit map applet with a static one that shows the number of visits per area of the sky for a given date range and instrument.


alserene (approved):
Thanks for all your efforts making these work and be pretty for the ND 🤩💛

PR #134, opened 2026-06-15
OSW-2117: Add Observatory Status timeline to the Context Feed.

alserene:
This work was done primarily by Owen, with Alice taking over when he went on leave.

There are a number of aspects that, due to our current time-pressures, might end up being implemented in a follow-up PR, such as tests. See [OSW-2444](https://rubinobs.atlassian.net/browse/OSW-2444) for the ticket tracking these tasks.

Example:
obs-status-timeline

Tooltip:
obs-status-tooltip

[OSW-2444]: https://rubinobs.atlassian.net/browse/OSW-2444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/pages/ContextFeed.jsx
```diff
@@ -309,16 +380,41 @@ function ContextFeed() {
...
+        event_time_dt: timeUTC,
+        event_time_millis: entry.time_ms,
+        event_type: "Observatory Status",
+        event_color: "#ffffff",
+        displayIndex: null,
+        name: getStateChangeDescription(obsStatusEntries, i),
+        description: entry.note ?? "",
+      };
+    });
+  }, [obsStatusEntries]);
```
alserene: @nemovc, obs status rows are already included in the fetched context feed data. Instead of creating our own rows, we just need to add an extra category (`23`) to `context-feed-definitions.js`.

While you're modifying that file, you can also rename and move to `/constants`.

src/components/context-feed-definitions.js
```diff
@@ -9,10 +9,11 @@ export const CATEGORY_INDEX_INFO = {
...
-  10: { color: "#ffffff", label: "AUTOLOG (Simonyi)", displayIndex: null },
-  11: { color: "#ffffff", label: "AUTOLOG (AuxTel)", displayIndex: null },
-  12: { color: "#ffffff", label: "AUTOLOG (General)", displayIndex: null },
+  10: { color: "#a8a29e", label: "AUTOLOG (Simonyi)", displayIndex: null },
+  11: { color: "#a8a29e", label: "AUTOLOG (AuxTel)", displayIndex: null },
+  12: { color: "#a8a29e", label: "AUTOLOG (General)", displayIndex: null },
   20: { color: "#14b8a6", label: "Narrative Log (Simonyi)", displayIndex: 9 },
   21: { color: "#38bdf8", label: "Narrative Log (AuxTel)", displayIndex: 10 },
   22: { color: "#3b82f6", label: "Narrative Log (General)", displayIndex: 11 },
+  30: { color: "#ffffff", label: "Observatory Status", displayIndex: null },
```
alserene: See comment in `/ContextFeed.jsx`, this should be `23: { ... }`.

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
```diff
@@ -0,0 +1,136 @@
...
+  DIST_BELOW_X_AXIS: 16,
+
+  // Text styling
+  LABEL_TEXT_SIZE: 14,
+  AXIS_LABEL_TEXT_SIZE: 12,
+
+  // Bar/marker styling
+  BAR_HEIGHT: 12,
+  MARKER_SIZE: 14,
+  MARKER_SIZE_WITH_NOTE: 16,
```
alserene, line 73: Change property names to camelCase pls.

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
```diff
@@ -0,0 +1,136 @@
...
+ * Twilight line styling
+ */
+export const TWILIGHT_STYLES = {
+  TWELVE_DEG: {
+    color: "#0ea5e9", // Sky blue
+    lineWidth: 2,
+    lineType: "solid",
+    label: "12°",
+  },
+  ZERO_DEG: {
```
alserene: Change property names to camelCase pls (`TWELVE_DEG` -> `twelveDeg` and `ZERO_DEG` -> `zeroDeg`).

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
```diff
@@ -0,0 +1,136 @@
...
+  },
+};
+
+/**
+ * Opacity values for active/inactive states
+ */
+export const STATUS_OPACITY = {
+  ACTIVE: 1,
+  INACTIVE: 0.3,
+  HOVER: 0.8,
```
alserene: Change property names to camelCase pls.

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+/**
+ * Color scheme for each observatory state
+ * Colors chosen for visual distinction and semantic meaning
+ */
+export const STATUS_COLORS = {
+  UNKNOWN: "#0C4A47",
+  DAYTIME: "#22C55E",
+  OPERATIONAL: "#22C55E",
```
alserene: Pls change colours to match Figma :)

```suggestion
  DAYTIME: "#00A370",
  OPERATIONAL: "#00A370",
```

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
```diff
@@ -0,0 +1,136 @@
...
+};
+
+/**
+ * Bar colors at 40% intensity (blended with black background)
+ * These are solid colors that appear as if the STATUS_COLORS are at 40% opacity
+ */
+export const STATUS_BAR_COLORS = {
+  UNKNOWN: "#051E1C",
+  DAYTIME: "#0E4F26",
+  OPERATIONAL: "#0E4F26",
```
alserene: Pls change colours to match Figma :)

```suggestion
  DAYTIME: "#00412D",
  OPERATIONAL: "#00412D",
```

src/utils/observatoryStatusUtils.js
```diff
@@ -0,0 +1,226 @@
...
+ * @param {Array} entries - Array of status entries from API, sorted by time:
+ *   [{ time: string, status: number, note: string, statusLabels: string, time_ms: number }]
+ * @param {number} endTime - End time of the visible range (to close final intervals)
+ * @returns {Object} Object with state names as keys and arrays of intervals as values:
+ *   {
+ *     UNKNOWN: [{ start, end, note, time }],
+ *     DAYTIME: [{ start, end, note, time }],
+ *     ...
+ *   }
+ */
```
alserene: This docstring has been detached from its function, `transformStatusToSeries`. Please reunite them 🙏

src/utils/timeUtils.js
```diff
@@ -285,6 +285,35 @@ function utcDateToCalendarDate(utcDate) {
...
+ * Formats a timestamp (milliseconds since epoch) to the standard context feed table format.
+ *
+ * @param {number} millis - Timestamp in milliseconds since epoch.
+ * @returns {string} Formatted date-time string as "yyyy-LL-dd HH:mm:ss.S"
+ */
+function formatTimestamp(millis) {
+  return DateTime.fromMillis(millis, { zone: "utc" }).toFormat(
+    "yyyy-LL-dd HH:mm:ss.S",
+  );
+}
```
alserene, line 310: This function already exists in `ContextFeedColumns.jsx` (see [here](https://github.com/lsst-ts/ts_logging_frontend/blob/ec3f3f4a69b838255e75d96218d4d0614bb2f435/src/components/ContextFeedColumns.jsx#L27)), so we should combine these into one. The other one is more general (accepting timezone) and has a fallback, so maybe we should use that one?

alserene:
I approve this, provided the requested changes are made and the existing failing (commented out) tests are fixed. Also the addition of unit tests for the utils would be good. E2E tests would of course be good for the timeline component, but we don't yet have CF E2E tests, so maybe that is its own PR...

Anyway, I can't actually approve due to me making the PR. So... @nemovc, once you make ze changes, you can approve it on my behalf 🙃

src/components/ObservatoryStatusTimeline.jsx
```diff
@@ -0,0 +1,486 @@
...
+import {
+  TIMELINE_MARGINS,
+  TIMELINE_DIMENSIONS,
+} from "@/constants/TIMELINE_DEFINITIONS";
+import { useEChartsTimeline } from "@/hooks/useEChartsTimeline";
+
+/**
+ * Observatory status timeline component.
+ *
+ * Displays 7 parallel series showing when each observatory state was active,
```
alserene: `Displays` -> `Display`. Docstrings, in general, I'm pretty sure, should be in imperative tense...

tests/e2e/mocked/plots/almanac.spec.js
```diff
@@ -56,18 +56,18 @@ test.describe("Twilight lines — full night range", () => {
...
+  // }) => {
+  //   const charts = page.locator("[data-slot='chart']");
+  //   await expect(charts).toHaveCount(4);
+
+  //   for (let i = 0; i < 4; i++) {
+  //     await expect(
+  //       charts.nth(i).locator(`line[stroke="${TWILIGHT_STROKE}"]`),
+  //     ).toHaveCount(2);
+  //   }
+  // });
```
alserene: Make them pass pls 🙏

alserene: x6

alserene:
Hehe, you saying that there haven't been any noticeable problems on dev reminded me that I saw that the tooltip can get cut off when your window width is small:

tooltip cut off

To fix, just add `confine: true,` to `option.tooltip`. As far as I can tell, it only makes the tooltip cover over the point of hover in the instances where the tooltip would be cut off.

src/pages/ContextFeed.jsx
nemovc: So we don't actually need to fetch the obs-status endpoint at all?

This also means that the obs-status timeline will no longer load faster than the main timeline, which is probably fine I guess

nemovc: Nah, we need the endpoint, since the CF rows do not contain the status as an int

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
nemovc: disagree, these are constants, even if they're "constants inside an object"

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
nemovc: disagree for same reason as above, these are constants. The inner property names being camelCase is because the 'constant property' is TWELVE_DEG

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
nemovc: see above

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
nemovc: I have no idea how this changed, so thank you for this good catch

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
nemovc: TIMELINE_DEFINITIONS.js is also SHOUTING_CASE for all its constants

alserene: You'll have to fight my LLM on this.

src/utils/timeUtils.js
nemovc: Not quite the same, since this one just accepts int milliseconds while the ContextFeedColumns one accept any old thing (String? Date object? Number? null? doesn't matter, it'll produce a cell). But you're right that there shouldn't be two ways to format a timestamp, so I've combined them

nemovc (approved):

PR #135, opened 2026-06-18
OSW-2330 Time Loss: add Computed Fault to new card.

emanehab99:
Add calculated fault loss to Time Loss card.
Also refactor Time Accounting applet
This [backend PR](https://github.com/lsst-ts/ts_logging_and_reporting/pull/130) is needed for this one to work
Screenshot from 2026-06-17 18-33-24

Screenshot from 2026-06-18 11-52-33

### DIFF
--- a/src/components/TimeAccountingApplet.jsx
+++ b/src/components/TimeAccountingApplet.jsx
@@ -11,7 +11,6 @@ import WarningIcon from "../assets/WarningIcon";
 import { Skeleton } from "@/components/ui/skeleton";
 import { ChartContainer, ChartTooltip } from "@/components/ui/chart";
 import { Cell, Bar, BarChart, XAxis, YAxis } from "recharts";
-import { DateTime } from "luxon";
 
 /**
  * Time accounting logic for night analysis.
@@ -27,30 +26,32 @@ function TimeAccountingApplet({
   loading,
   onSkyTimeAccounting,
   sumOnSkyExpTime,
-  nightHours,
+  elapsedTwilightHours,
   closedDomeHours,
   calculatedFaultHours,
   faultDataUnavailable,
   faultErrorMessage,
+  domeError,
 }) {
   // console.log(`openDomeError: ${openDomeError}`);
   // console.log(`timeAccountingError: ${timeAccountingError}`);
   // const timeAccountingUnavailable = Boolean(timeAccountingError);
-  // const domeUnavailable = Boolean(openDomeError);
+  const domeUnavailable = Boolean(domeError);
   // Fault is derived from overhead (time accounting) AND weather/dome-closed
   // loss (dome data), so it's unreliable if either upstream source failed.
   // const faultUnavailable = timeAccountingUnavailable || domeUnavailable;
 
   const [expPercent, nonExpPercent] = useMemo(() => {
-    if (!nightHours || nightHours === 0) {
+    if (!elapsedTwilightHours || elapsedTwilightHours === 0) {
       return [0, 0];
     }
     const expHours = sumOnSkyExpTime / 3600;
-    // calculate percentage of on-sky exposure time of night hours between 12 deg twilights
-    const expPercentage = Math.round((expHours / nightHours) * 100);
+    // Calculate the percentage of on-sky exposure time relative to the
+    // completed 12-degree twilight hours in the selected range.
+    const expPercentage = Math.round((expHours / elapsedTwilightHours) * 100);
     const nonExpPercentage = 100 - expPercentage;
     return [expPercentage, nonExpPercentage];
-  }, [sumOnSkyExpTime, nightHours]);
+  }, [elapsedTwilightHours, sumOnSkyExpTime]);
 
   // One hue (teal), varying only in lightness, for the four bars that are
   // sub-components of the same two metrics (gap time, overhead time) split
@@ -91,6 +92,7 @@ function TimeAccountingApplet({
   // TODO: Test with a night with only non-science exposures and the dome was open 20260615
   // TODO: Test with a night in the future
   // TODO: check error in night 20260614
+
   const chartData = [
     {
       name: "Gaps",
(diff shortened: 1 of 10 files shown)


src/components/TimeAccountingApplet.jsx
```diff
@@ -24,279 +23,50 @@ import { almanacDayobsForPlot, TAI_OFFSET_SECONDS } from "../utils/timeUtils";
...
-      gapWithFilterChange: filterChangeHours,
-      gapWithoutFilterChange: noFilterChangeHours,
-      overheadWithFilterChange: overheadFilterChangeHours,
-      overheadWithoutFilterChange: overheadNoFilterChangeHours,
-      domeClosedHours: domeClosed,
-      calculatedFaultTime: faultTimeHours,
-    };
-  }, [groupedByDayobs, nights, weatherLossHours, currentNight]);
+  console.log(`openDomeError: ${openDomeError}`);
+  console.log(`timeAccountingError: ${timeAccountingError}`);
```
alserene: Pls remove :)

src/pages/Digest.jsx
```diff
@@ -128,7 +132,7 @@ export default function Digest() {
     setSumOnSkyExpTime(0.0);
     setSumExpTime(0);
     setJiraTickets([]);
-    setNarrativeWeatherLoss(0.0);
+    // setNarrativeWeatherLoss(0.0);
```
alserene: We should add a TODO here with the ticket number for the weather ticket.

src/utils/utils.js
```diff
@@ -577,6 +577,40 @@ const getBlockSourceLabel = (source) =>
...
+) => {
+  const expTimeHours = onSkyExpTime / 3600;
+
+  // console.log(`openDomeHours: ${openDomeHours}`);
+  console.log(`nightHours: ${nightHours}`);
+  // console.log(`openHours: ${domeTotals.openHours}`);
+  console.log(`closedHours: ${closedDomeHours}`);
+  console.log(`expTimeHours: ${expTimeHours}`);
+  console.log(`timeAccounting: ${onSkyTimeAccounting}`);
+  console.log(`weatherLoss: ${weatherLossHours}`);
```
alserene: Pls remove :)

src/utils/utils.js
```diff
@@ -577,6 +577,40 @@ const getBlockSourceLabel = (source) =>
...
+
+  const faultTimeHours =
+    nightHours -
+    expTimeHours -
+    onSkyTimeAccounting.sum_overhead_without_filter_change -
+    onSkyTimeAccounting.sum_overhead_with_filter_change -
+    Math.min(closedDomeHours, weatherLossHours ?? 0);
+
+  console.log(`faultTimeHours: ${faultTimeHours.toFixed(2)}`);
```
alserene: Pls remove :)

src/pages/Digest.jsx
```diff
@@ -155,15 +164,24 @@ export default function Digest() {
...
           setOnSkyExpCount(onSkyExpNo);
           setSumOnSkyExpTime(totalOnSkyExpTime);
           setExposuresLoading(false);
-          setOpenDomeTimes(openDomeTimes);
+          setOpenDomeTimes(domeTimes);
+          setDayObsOpenDomeHours(openDomeHours);
+          setOnSkyTimeAccounting(nightOnSkyTimeAccounting);
+          setOpenDomeError(domeError);
+          setTimeAccountingError(accountingError);
+          console.log(openDomeTimes);
```
alserene, line 187: remove pls :)

src/pages/Digest.jsx
```diff
@@ -231,9 +261,9 @@ export default function Digest() {
       });
 
     fetchNarrativeLog(startDayobs, queryEndDayobs, instrument, abortController)
-      .then(([weather, fault]) => {
-        setNarrativeWeatherLoss(weather);
-        setNarrativeFaultLoss(fault);
+      .then((data) => {
+        // setNarrativeWeatherLoss(weather);
+        setNarrativeFaultLoss(data.time_lost_to_faults);
```
alserene, line 267: In the new dev release, fault loss is showing as `NA` because of this change.

**Cause**
The endpoint returns a `dict` and then the fetch util converts that to a `list`, so the reference here needs to be order-based, not name-based (_unless_ we change the fetch, and that is whole other conversation, because lots of our fetch utils return ordered `lists` rather than structured `dicts`, and this should probably change).

Backend, `main.py`:

```
return {
    "narrative_log": records,
    "time_lost_to_weather": time_lost_to_weather,
    "time_lost_to_faults": time_lost_to_faults,
}
```

Frontend, `fetchUtil.jsx`:

```
return [
  data.time_lost_to_weather,
  data.time_lost_to_faults,
  data.narrative_log,
];
```

So, please change to:

```
.then(([_, fault]) => {
        setNarrativeFaultLoss(fault);
```

emanehab99:
I changed the PR to `draft` while I'm adding tests and trying it using `rubin-nights` to `0.18.0` where the open dome times changes are introduced

src/components/TimeAccountingApplet.jsx
```diff
@@ -24,279 +22,43 @@ import { almanacDayobsForPlot, TAI_OFFSET_SECONDS } from "../utils/timeUtils";
...
+  // by filter change
+  const COLOR_GAPS = "#5eead4"; // teal-300
+  const COLOR_OVERHEAD = "#14b8a6"; // teal-500
+  const COLOR_GAPS_FILTER_CHANGE = "#0f766e"; // teal-700
+  const COLOR_OVERHEAD_FILTER_CHANGE = "#134e4a"; // teal-900, matches CardHeader
+
+  // Fault and Closed Dome come from different upstream sources and are
+  // genuinely distinct categories, so they get their own hues
+  const COLOR_FAULT = "#dc2626";
+  const COLOR_CLOSED_DOME = "#71717a";
```
alserene, line 61: I don't really understand when we should define constants within the file they're used or within some `constants/definitions.js` file.

src/utils/utils.js
```diff
@@ -577,6 +575,55 @@ const getBlockSourceLabel = (source) =>
...
 
+/**
+ * Check whether a plain object has no own enumerable properties.
+ *
+ * @param {Object|null|undefined} obj - Object to test.
+ * @returns {boolean} `true` when `obj` is an empty plain object.
+ */
+const isDictionaryEmpty = (obj) => {
+  return obj && Object.keys(obj).length === 0 && obj.constructor === Object;
+};
```
alserene, line 586: Ooo I like this! I feel like there are definitely other places that this can be used. I'll keep an eye out :)

src/components/TimeAccountingApplet.jsx
alserene: Nothing to change here, just a comment 🙃

src/pages/Digest.jsx
alserene: @emanehab99 , did this get fixed?

alserene (approved):
Thanks, Eman :)

src/pages/Digest.jsx
emanehab99: I fixed the fetch utility to return the `data` object instead of a list to match the call in the `Digest` page.

src/components/TimeAccountingApplet.jsx
emanehab99: If there is a possibility the constant can be used somewhere else, like band colors, then moving it to a separate file makes sense.

PR #137, opened 2026-06-26
OSW-2118: Replace Time Accounting applet with Observatory Status cumulative plot

alserene:
This PR welcomes a new applet to our Digest page, the Observatory Status cumulative plot.
- It swaps out the old Time Accounting applet for a cumulative plot using Observatory Status data.
- Some new features/implementations, including a dialog to full-screen and some native eCharts zooming functionality that differs to our usual implementation.


emanehab99 (approved):
Great work on this PR, Alice! 👏 🎉  This is a complex feature, and you’ve done a great job handling the different behaviors and edge cases. The unit test coverage is also really strong.

Thanks for putting together a detailed ticket description. Even though a few details have changed along the way, having that level of context made the review much easier.

There’s really only one behavior that stood out to me from a UX perspective: handling very long notes in the plot tooltip. It’s useful that the user can access the full note, but with very long text the tooltip can cause a significant change in page height and quite a bit of page movement while hovering around the plot. Scrolling down to read the note can also be difficult because the point may move out of mouse focus, which closes the tooltip. I’ve attached a screenshot showing an example.

Did you consider any alternative ways of handling longer notes? One option could be to keep the hover tooltip short by truncating the note, and then allow the user to click the point to open the full text in a [Sheet](https://ui.shadcn.com/docs/components/base/sheet).

I think this might need a bit of UX exploration rather than rushing to resolve it in this PR. We could create a follow-up ticket to look at the behavior and possible alternatives separately.

A couple of other things I noticed:

* The tooltip styling seems slightly different across the three plots in the Digest. Would it make sense to make these consistent?
* I noticed a few behaviors aren’t currently covered by the e2e tests: the zoom interactions (rectangle/LineX zooming and double-click reset), multi-night plots, legend toggling, and partial data availability. Would some of them be worth covering?

Image

alserene:
Thanks, @emanehab99, for your review 😊

**Loooooong notes**
Yes, I noticed this tooltip issue, also, and have been thinking about it. These extremely long notes are caused by a LOVE input bug where the notes are getting duplicated across successive entries, so hopefully they will stop happening once that gets fixed. But, yes, we do need to guard for them, especially because (I think) the new note gets appended to the end! I will look into this and your idea of a sheet, thanks!

**Tooltip styling**
The tooltip style differences, yes, I agree. I copied here the tooltip styling for the CF obs-status timeline, but that was perhaps not the correct strategy. I think at this stage we have 3x different tooltips because the plots are different, too 🙈 Not good!

**E2E tests**
For the testing, I wanted to test all of those things you suggested, but consultation with the agent suggested they were either too much stepping-on-ECharts-toes sort of thing (e.g. the legend testing) or too difficult/flaky (the zooming), so I left it out for now. I will reconsider and see if there are other ways to test this, e.g. with a screenshot test (although, I know Asher disapproves of these).

src/components/ObservatoryStatusApplet.jsx
```diff
@@ -0,0 +1,179 @@
...
+  }.`;
+
+  return (
+    <Card className="border-none p-0 bg-stone-800 gap-2">
+      <CardHeader className="grid-cols-3 bg-teal-900 p-4 rounded-sm align-center gap-0">
+        <CardTitle className="text-white font-thin col-span-2">
+          <div className="flex flex-row gap-1">
+            <p>Observatory Status</p>
+            {/* Partial data availability warning */}
+            {!loading && obsAvailabilityStatus === "partial" && (
```
sebastian-aranda: It would be good to define somewhere the possible values that can come from the backend (at least the ones used for conditions). So far I found "`none`" and "`partial`" in the code.

```suggestion
            {!loading && obsAvailabilityStatus === OBS_STATUS_AVAILABILITY_STATUSES.PARTIAL && (
```

src/components/ObservatoryStatusCumulativePlot.jsx
```diff
@@ -0,0 +1,584 @@
...
+    // because there's no toolbox button to enable it.
+    instance.dispatchAction({
+      type: "takeGlobalCursor",
+      key: "brush",
+      brushOption: { brushType: "rect", brushMode: "single" },
+    });
+
+    // Defer graphic elements until ECharts has finished rendering,
+    // since convertToPixel is only valid after the chart is laid out.
+    setTimeout(() => updateGraphicElements(instance), 0);
```
sebastian-aranda, line 216: Nice 💡 I didn't know this trick.

src/components/ObservatoryStatusCumulativePlot.jsx
```diff
@@ -0,0 +1,584 @@
...
+    yDomain,
+    updateGraphicElements,
+    instanceRef,
+  ]);
+
+  // Re-run graphic elements when the relevant props change
+  useEffect(() => {
+    const instance = instanceRef.current;
+    if (instance) updateGraphicElements(instance);
+  }, [updateGraphicElements, instanceRef]);
```
sebastian-aranda: The previous effect also has `[updateGraphicElements, instanceRef]` on its dependency array and it also calls `updateGraphicElements(instance)` at the end (with a `setTimeout` though). Do we really need another useEffect that is triggering the same final effect?

src/components/ObservatoryStatusCumulativePlot.jsx
```diff
@@ -0,0 +1,584 @@
...
+          symbol: (value, params) =>
+            params.data.hasNote ? "diamond" : "circle",
+          symbolSize: (value, params) =>
+            params.data.hasNote
+              ? markerSize.WITH_NOTE
+              : markerSize.WITHOUT_NOTE,
+          z: STATUS_CUMULATIVE_ELEMENTS_Z.MARKERS,
+        })),
+      ],
+    };
```
sebastian-aranda: Moving this definition outside the component or extracting some sections can improve readability of this component. You could move out anything that is not component prop/state dependant.

e.g. moving the `formatter` and `position` method definitions out or maybe just moving the full option definition to something like: `const option = buildEChartSchema(...)`

src/constants/OBSERVATORY_STATUS_DEFINITIONS.js
sebastian-aranda, line 1: 👏

src/utils/timeUtils.js
```diff
@@ -326,6 +326,18 @@ function formatDuration(durationMs) {
   )}:${String(seconds).padStart(2, "0")}.${String(millis).padStart(3, "0")}`;
 }
 
+/**
+ * Returns the current dayobs as a formatted dayobs string.
+ *
+ * @param {string} [format="yyyy-LL-dd"] - The format string (e.g., "yyyy-LL-dd" or "yyyyLLdd").
```
sebastian-aranda:
```suggestion
 * @param {string} [format="yyyy-LL-dd"] - The luxon format string (e.g., "yyyy-LL-dd" or "yyyyLLdd"). See https://moment.github.io/luxon/#/formatting?id=table-of-tokens.
```

src/utils/timeUtils.js
```diff
@@ -326,6 +326,18 @@ function formatDuration(durationMs) {
...
+ * @returns {string} The current dayobs, as a formatted string.
+ */
+function getCurrentDayObs(format = "yyyyLLdd") {
+  return Number(
+    DateTime.now().setZone("UTC").minus({ hours: 12 }).toFormat(format),
+  );
+}
```
sebastian-aranda: This method is supposed to return a formatted string, but the function is parsing with `Number`. I suspect this will only work when the format is `yyyyLLdd`.

src/hooks/useEChartsBrushZoom.js
```diff
@@ -0,0 +1,157 @@
...
+      brushOption: {
+        brushType,
+        brushMode: "single",
+      },
+    });
+  };
+
+  const modifierKeys = useRef({
+    shift: false,
+  });
```
sebastian-aranda: Maybe group all `useRef` above, at the beginning of the component?

src/hooks/useEChartsBrushZoom.js
```diff
@@ -0,0 +1,157 @@
...
+      updateBrushType();
+    };
+
+    window.addEventListener("keydown", down);
+    window.addEventListener("keyup", up);
+
+    return () => {
+      window.removeEventListener("keydown", down);
+      window.removeEventListener("keyup", up);
+    };
```
sebastian-aranda: Both `down` and `up` methods are the same.
```suggestion
    const keyEvent = (e) => {
      modifierKeys.current.shift = e.shiftKey;
      updateBrushType();
    };

    window.addEventListener("keydown", keyEvent);
    window.addEventListener("keyup", keyEvent);

    return () => {
      window.removeEventListener("keydown", keyEvent);
      window.removeEventListener("keyup", keyEvent);
    };
```

src/hooks/useEChartsBrushZoom.js
```diff
@@ -0,0 +1,157 @@
...
+  }, []);
+
+  // Keep refs to latest prop values for use inside stable event handlers.
+  const onResizeRef = useRef(onResize);
+  useEffect(() => {
+    onResizeRef.current = onResize;
+  }, [onResize]);
```
sebastian-aranda: Do you really need a ref for this? Wouldn't just using onResize work fine?

src/pages/Digest.jsx
```diff
@@ -62,18 +62,17 @@ export default function Digest() {
...
-  // const [obsStatusIntervals, setObsStatusIntervals] = useState([]);
+  const [obsStatusIntervals, setObsStatusIntervals] = useState([]);
   const [obsStatusFaultLoss, setObsStatusFaultLoss] = useState(0.0);
   const [obsStatusWeatherLoss, setObsStatusWeatherLoss] = useState(0.0);
   const [obsStatusAvailability, setObsStatusAvailability] = useState(
     EMPTY_OBS_STATUS_AVAILABILITY,
   );
 
-  const [exposuresLoading, setExposuresLoading] = useState(false);
+  const [exposuresLoading, setExposuresLoading] = useState(false); // true or false?
```
sebastian-aranda: I would remove this comment unless this is producing a buggy behavior, in that case please create a ticket and reference it here, e.g.

```suggestion
  const [exposuresLoading, setExposuresLoading] = useState(false); // TODO: define if true or false. See OSW-XXXX.
```

sebastian-aranda (approved):
Amazing job @alserene 👏! thank you so much. I left some minor comments for your consideration, but I'm marking this as approved 👍 .

PR #141, opened 2026-07-31
OSW-2754: DataTable cell expansion bugfix

nemovc:
The implementation of [row memoization](https://github.com/lsst-ts/ts_logging_frontend/commit/7e76e053e127f87bb6632a7b10781140ac5c320f) for the DataTable component contained a critical flaw, which meant that traceback and YAML cells could not be expanded.

This PR moves cell expansion state into the individual cells themselves, so that the cell state doesn't have to be stored at the table level and passed through to each cell. More importantly, it widens the list of properties that the cell memoizer considers when checking whether a cell needs an update.

My investigation also showed that the DataTable `useURLSync` hook was firing incorrectly (and leading to page rerenders) when the `selected` query parameter was changed - this is now narrowed following the same pattern used in [other similar hooks](https://github.com/lsst-ts/ts_logging_frontend/commit/77112a5e5491908ebcd1ffddb6755b8a8bc13f2e). This also added some memoization/reference stability, which I expanded to also cover the default columns paths. 

A new bug was found which I fixed immediately as it's of similar importance, but happy to move into a separate ticket.
Filtering on non-querystring columns was not working (e.g., you could select a filter for "Final Status" in the Context Feed, but nothing would happen). This was due to the `useURLSync` hook silently dropping columns it didn't know about, and is fixed in this PR with the addition of local state to the hook


src/components/DataLogTable.jsx
```diff
@@ -40,7 +40,7 @@ function DataLogTable({
   return (
     <DataTable
       data={data}
-      columns={dataLogColumns[telescope] ?? []}
+      columns={dataLogColumns[telescope]}
```
alserene, line 43: Why did you remove the fallbacks?

alserene (approved):
Nice one 😎

- I tested locally and the bugs are fixed and I couldn't see any new ones 👍
- I think I follow the code changes. I commented one question about removing fallbacks.

I did notice an existing undesirable behaviour that we probably want to change, and if its not too difficult a fix we could pop it in here? If not, lets ticket it. When we expand any traceback/yaml, I would think the `Collapse All Tracebacks`/`Collapse All YAML` boxes should become unticked. Because, if you have some cells expanded, and then want to collapse them all, you have to untick and then tick the checkbox.

src/components/DataLogTable.jsx
nemovc: because it was defensive programming that literally couldn't be reached. `telescope` is always a valid key of `dataLogColumns`

alserene:
Just a note here that the Data Log E2E Tests PR ([OSW-2155](https://github.com/lsst-ts/ts_logging_frontend/pull/127)) has some changes that conflict with this PR.

[OSW-2155]: https://rubinobs.atlassian.net/browse/OSW-2155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

nemovc:
> I did notice an existing undesirable behaviour that we probably want to change, and if its not too difficult a fix we could pop it in here? If not, lets ticket it. When we expand any traceback/yaml, I would think the `Collapse All Tracebacks`/`Collapse All YAML` boxes should become unticked. Because, if you have some cells expanded, and then want to collapse them all, you have to untick and then tick the checkbox.

I think this should be its own ticket, just so we can land this PR as is. 

While I agree that there's a disconnect between the state of the overarching checkbox and the per-row expansion state, I think the base issue is that fundamentally a two-state checkbox is not the correct control to use here. So we should probably discuss that as well

src/pages/ContextFeed.jsx
```diff
@@ -113,7 +113,14 @@ function ContextFeed() {
...
   // Default event type filters based on telescope
-  const defaultEventTypes = filterDefaultEventsByTelescope(telescope);
+  const defaultEventTypes = useMemo(
+    () => filterDefaultEventsByTelescope(telescope),
+    [telescope],
+  );
+  const defaultFilters = useMemo(
+    () => [{ id: "event_type", value: defaultEventTypes }],
+    [defaultEventTypes],
+  );
```
sebastian-aranda, line 123:
```suggestion
  const defaultFilters = useMemo(
    () => [{ id: "event_type", value: filterDefaultEventsByTelescope(telescope) }],
    [telescope],
  );
```

sebastian-aranda (commented):
Agh, I feel this memoization attempt has gotten a bit out of hand. I really like the simplification of moving the expanded state handling into the child components (Cells), but I really don't like that `resetToken` for triggering renders and we still cannot avoid to be tied to the "table-level-state". Having said that, this works anyways, but I'm not happy with the extra complexity of _memoizing_ TableRow (introduced in the previous PR) as I feel the current props comparison is not a silver bullet and that could lead to premature memoizations due nested table props that should triger renders. My main ask/suggestion in that sense would be to just remove the memoization of TableRow and leave the optimization for a future attempt by checking into virtualization technologies e.g. https://tanstack.com/virtual/latest. Regarding the `resetToken` I can live with that, although I would recommend to just go back to the original implementation with the expanded state and setters lifted into the parent "table-level-state", but my happiness won't be affected if you disagree with this :grin:. Please consider removing the memoization though.

The rest looks pretty good, thanks for the `useURLSync` catch :raised_hands: !

sebastian-aranda (approved):

nemovc:
I KNEW I solved the 'can't filter some columns' bug - it's in the data-log tests branch, which is why I'd forgotten since that branch pre-dates the human use of fire.`

PR #142, opened 2026-08-06
OSW-2773: Implement automatic CI pipeline for image building in merges to develop.

sebastian-aranda:
This PR refactors the `Jenkinsfile` to have custom steps for image building when merging PRs to `develop`. A new `Dockerfile-deploy` file was added with the specifications for building images for our deployments.

Additionally adjustments were required in the section where we display packages version. As automatic builds won't create a git tag, this affects the current way we are showing versions. I would like to keep git tags only for manual processes (prod release or alpha manual release for dev), so we keep control of the set of changes we are packaging (instead of being creating tags for every single merge to develop).

Also the `make_release.py` script has been extended to support configuring which steps to run. Used for the automatic image build process and package.json version bumps.

See https://github.com/lsst-ts/ts_logging_and_reporting/pull/140.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,37 +1,24 @@
-properties(
-    [
-    buildDiscarder(
-        logRotator(
-            artifactDaysToKeepStr: '',
-            artifactNumToKeepStr: '',
-            daysToKeepStr: '14',
-            numToKeepStr: '10',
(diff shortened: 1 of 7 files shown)


Vebop (approved):
Thanks for making these updates!

